Issue #29, opened 2021-08-16
add label/reason/solution to the response


lsmith77:
this will require integrating translation.io data

PR #33, opened 2021-08-17
add translations for reason/solution

lsmith77:
relates to https://github.com/witty-works/nlp_api/issues/29


lsmith77:
@lnazarenko I am just adding the principle ability to translate content here. for the actual translations we will need the proper category, which we only have in the POS API. so I guess we can merge it as is for now and then look to do the proper translation when we merge the entities/POS API on Thursday.

Also I then have to get those translations into the .po format from translation.io, checking with the developers to see if that is possible.

PR #38, opened 2021-08-23
Update main.py

lnazarenko:
Add reasons&solutions for output

PR #39, opened 2021-08-23
refactored check API

lsmith77:
inspired by https://github.com/witty-works/nlp_api/pull/38/files I made the following changes.

note the `False Positive` category was just a quick fix. We don't actually want to include false positives in the response but this was the quickest way to keep the business logic.

I also removed all the functions that are not being called.

I also made some minor changes to the model, which we need to discuss with @arnaumanyosa


lnazarenko (approved):

lnazarenko (commented):
There is no "lang" parameter for the MaleCodedWordAnalysis function as an input compare to other function. Do you forgot it or other reasons for this?

PR #43, opened 2021-08-24
create a class to handle the language detection and translation

lsmith77:
I created a class that combines all the language/translation stuff. making it easier to pass the relevant data around


lnazarenko (approved):
Looks nice

PR #52, opened 2021-08-27
Integrate Blackfire as FastAPI middleware

sumerc:
Integrate Blackfire as middleware.


.github/workflows/main.yaml
```diff
@@ -25,9 +25,10 @@ jobs:
           key: ${{ runner.os }}-pipenv-${{ hashFiles('**/Pipfile.lock') }}
 
       - name: Install dependencies
-        if: steps.cache-pipenv.outputs.cache-hit != 'true'
```
lsmith77: do we still need to comment this?

app/main.py
```diff
@@ -73,6 +73,9 @@
     allow_headers=["*"],
 )
 
+from app.middleware import BlackfireFastAPIMiddleware
```
lsmith77, line 83: thx to this middlware we don't need to use `blackfire-python`, correct?

.github/workflows/main.yaml
lsmith77: otherwise we can move the install of the whl to a separate step if we always need to do this and cannot cache

app/main.py
sumerc: Yes. For now, we do this manually but once we officially support FastAPI, `blackfire-python` will auto-instrument the code.

.github/workflows/main.yaml
sumerc: You are right. I left that intentionally for you to test. Let me try your suggestion.

lsmith77 (approved):

PR #90, opened 2021-09-25
added languagetool support using aiohttp

### LINKED ISSUE #18: Integrate languagetool
* adding grammar/spelling
* allowing us to add custom rules based on grammar (like DB ist eine Arbeitgeberin)

lsmith77:
fixes #18 

you can try `Halo Welt` on https://pr-90-mbafagy-qh3uq7skrqxzg.de-2.platformsh.site/docs#/default/check_query_check_post

there is a competing PR in #91 but I think this approach is better as it gives us more control (which I used to make this async though I am not sure if the implementation is correct, see my comments).

I removed `response_lang` since this concept isn’t support in LanguageTool and it may also apply to us eventually since it may become too hard to maintain translations for each reason/solution for language specific rules

further reading: https://testdriven.io/blog/concurrency-parallelism-asyncio/
consider usin https://pypi.org/project/pytest-asyncio/

not sure if doing the following with both the LT call and the NLP work is the way to go:
```
async def main():
    tasks = []

    for i in range(5):
        tasks.append(write_genre(f"./async/new_file{i}.txt"))

    await asyncio.gather(*tasks)
```

but generally all we are doing here is helping concurrency, which doesn’t help the latency of individual requests, which may be achieved with parallelism via threading https://towardsdatascience.com/asyncio-is-not-parallelism-70bfed470489

https://medium.com/fintechexplained/advanced-python-concurrency-and-parallelism-82e378f26ced

should do some ApacheBench testing to figure out with which approach we achieve the best balance of latency while serving as many concurrent requests as possible

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -4,7 +4,7 @@ type: 'python:3.8'
 
 dependencies:
   python3:
-    pipenv: '2018.10.13'
+    pipenv: '2021.5.29'
     supervisor: '*'
 
 web:
@@ -23,6 +23,9 @@ mounts:
   'logs':
     source: local
     source_path: logs
+  'languagetool-server':
+    source: local
+    source_path: 'languagetool-server'
 
 disk: 4096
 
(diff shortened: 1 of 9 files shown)


.platform.app.yaml
```diff
@@ -4,7 +4,7 @@ type: 'python:3.8'
 
 dependencies:
   python3:
-    pipenv: '2018.10.13'
+    pipenv: '2021.5.29'
```
lsmith77, line 7: this is not relevant for the other changes

app/main.py
```diff
@@ -170,19 +172,19 @@ def get_root():
     return RedirectResponse(url='/form', status_code=301)
 
 @app.get("/docs", include_in_schema=False)
-async def get_swagger_documentation(username: str = Depends(get_current_username)):
+def get_swagger_documentation(username: str = Depends(get_current_username)):
```
lsmith77, line 175: lets only mark those endpoints as async that really leverage async capabilities

app/main.py
```diff
@@ -199,34 +201,82 @@ async def check_query(user_request_in: UserRequestIn, background_tasks: Backgrou
...
+                            None,
+                            match["shortMessage"],
+                            "",
+                            match["message"]
+                        )
+                    )
+
+    return list_results
+
+def language_rules(lang, text):
```
lsmith77, line 256: we might benefit here from using multiple threads .. but I think for now its ok like it is

PR #137, opened 2021-10-19
ensure that alternative case matches what they replace #136

lsmith77:
blocked by #135

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -40,8 +40,8 @@ def test_api():
         "start": 11,
         "end": 16,
         "alternatives": [
-            "jemand, der erfahren und fachkundig ist",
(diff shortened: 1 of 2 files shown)


tests/test_api.py
```diff
@@ -40,8 +40,8 @@ def test_api():
         "start": 11,
         "end": 16,
         "alternatives": [
-            "jemand, der erfahren und fachkundig ist",
-            "jemand mit Know-how und Ausdauer",
+            "Jemand, der erfahren und fachkundig ist",
```
lsmith77, line 43: the algorithm is a bit too naiv here but not sure if its still sensible to add this for now until we figure out a better approach.

lsmith77: @lnazarenko FYI

PR #140, opened 2021-10-19
127: Migrate to Pydantic settings management.

Szarad:
Use Pydantic Settings utility to handle environment variables as different types and for validation.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,19 @@
+from app.models import (
+    Config,
+    LangType,
+    Lang,
+    RequestIn,
+    RequestInEvent,
+    ResultOut,
+    ResultsOut,
+)
+import ast
+import re
+from spacy.tokens import Doc
+from spacy.matcher import PhraseMatcher, Matcher
+import spacy
+import pandas as pd
+from datetime import datetime
 import uvicorn
 import os
 import json
@@ -31,6 +47,7 @@ from fastapi.exception_handlers import (
 from typing import Optional
 from pydantic import BaseSettings
 
+
 class Settings(BaseSettings):
     """Load environment variables to python objects using pydantic."""
     logging_enabled: bool = False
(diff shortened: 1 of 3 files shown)


app/main.py
```diff
@@ -29,18 +29,20 @@
 )
 
 from typing import Optional
+from pydantic import BaseSettings
 
-os.environ["KMP_DUPLICATE_LIB_OK"] = "True"
```
Szarad, line 33: What this variables does? I couldn't find where it is used in the code, so I removed it. Let me know if it is used by some subprocess, then I will revert it.

lsmith77: based on https://stackoverflow.com/questions/53014306/error-15-initializing-libiomp5-dylib-but-found-libiomp5-dylib-already-initial I think this is something that OSX users should then just add to their local `.env` file.

in this spirit it would be good if you would also document how to create a `.env` file .. maybe you add a `.env.default` file into the repo or can pydantic generate such a file via some command?

FTR we have a small CLI script already here (maybe should be moved to a sub directory):
https://github.com/witty-works/nlp_api/blob/dev/update_locales.py

Szarad: In the end I've added `.env.development.mac` file.  I've another pull request with documentation and I explain there how to use this or setting env variable when activating a virtual environment. In Docker this variable will not be needed. I will squash commits when merging.

app/main.py
```diff
@@ -102,19 +104,19 @@ async def custom_http_exception_handler(request, e):
...
+if settings.languagetool_api:
+    languagetool_url = settings.languagetool_api
+elif settings.platform_relationships is not None:
+    relationships = json.loads(base64.b64decode(settings.platform_relationships))
     languagetool = relationships["languagetool"][0]
     languagetool_url = "%(scheme)s://%(host)s:%(port)d/v2" % languagetool
 
 security = HTTPBasic(auto_error=False)
 
 basic_auth = {
```
lsmith77: maybe we can kill this entire dict .. especially with the settings object we can just use the `settings…` in the relevant places and its easy to read.

lsmith77 (approved):

lsmith77 (approved):

PR #148, opened 2021-10-21
make the workers relative to the available memory

lsmith77:
The issue I am looking to fix here is that the dev envs don't have as much resources so we need to reduce the amount of workers. However I also want to generally make the application adjust the amount of workers relative to available resources (ie. if we switch to a bigger plan on platform.sh we launch more workers)

Unfortunately /run/config.json is hardcoded into the image so it doesn't change depending on if its the prod env or the dev env. However since we now that for dev env, its a size S plan (for prod we currently use a size L plan), we know the available resources and so I hardcoded the value there.

### DIFF
--- a/.platform.app.yaml
+++ b/.platform.app.yaml
@@ -16,7 +16,9 @@ web:
   upstream:
     socket_family: unix
   commands:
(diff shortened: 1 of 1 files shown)


Szarad (commented):
Changes look good for me, but could we discuss a bit platform.sh and how I can run this exact command in the platform?

Szarad (approved):

PR #150, opened 2021-10-22
134: build docker image dev

Szarad:


lsmith77:
ok cool .. will play with it.
could you also add instructions to the docs so that we don’t forget?

PR #161, opened 2021-10-27
add blackfire

lsmith77:


lsmith77:
@sumerc any update on the integration?

sumerc:
I think there were 2 issues on my side regarding integration:

1. Scenarios was not working with FastAPI. Today I am releasing a new probe version (1.8.0) that includes support for APM auto-profile + writing scenarios with `.blackfire.yml`. So, it should be fixed by simply downloading `1.8.0`. 
2. There is still an ongoing work on Python support for the Agent, but I think that does not stop us from moving forward as IIRC, the current branch was just working fine with some tweaks on env. vars..

Do you have any other issues in your mind?

Issue #171, opened 2021-11-05
Fallback Form


lsmith77:

lsmith77:
Link to feature: https://wittyworks.productboard.com/feature-board/planning/features/10509819

PR #172, opened 2021-11-05
remove mindbox infos from form #171

lsmith77:


Szarad (approved):
Changes look good, deployment works fine.

PR #173, opened 2021-11-08
134: Add Redis service to the platform.sh.

Szarad:
Add Redis service to dev environment.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,8 @@ babel = "*"
 aiohttp = "*"
 sentry-sdk = "*"
 opencensus-ext-azure = "*"
+redis = "*"
+platformshconfig = "*"
 
 [dev-packages]
 click = "*"
@@ -25,6 +27,8 @@ pytest-cov = "*"
 python-dotenv = "*"
 black = "*"
 opencensus-ext-azure = "*"
+pytest-asyncio = "*"
+fakeredis = "*"
 
 [requires]
 python_version = "3.9"
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,5 @@
+import platformshconfig
+from pydantic.types import Json
 from app.models import (
     Config,
     LangType,
@@ -6,6 +8,7 @@ from app.models import (
     RequestInEvent,
     ResultOut,
     ResultsOut,
+    ConfRequest,
 )
 import ast
 import re
@@ -58,6 +61,11 @@ import logging
 from opencensus.ext.azure.log_exporter import AzureLogHandler
 
 
+from redis import Redis
+import platformshconfig
+from fakeredis import FakeStrictRedis
+
+
 class Settings(BaseSettings):
     """Load environment variables to python objects using pydantic."""
 
@@ -70,6 +78,7 @@ class Settings(BaseSettings):
     api_docs_password: Optional[str]
     api_docs_auth_enabled: bool = False
     instrumentation_key: str = ""
+    testing: bool = False
 
     class Config:
         env_file = ".env"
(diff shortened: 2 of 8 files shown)


lsmith77:
@Szarad I think you are missing the change to `.platform/services.yml` to enable the redis service

Szarad:
What do you mean @lsmith77 ? The change is in .platform/services.yml and Redis service works (on PR env). 
Sorry for not putting it as draft,  I forgot.

lsmith77:
ah .. sorry .. misread. you are correct!

Szarad:
To do:
1. Add necessary libraries for tests not to fail.
2.  Add new unit tests
3. Change company<->user.id relationship
4. Optional: changes in the logic
5. Exception handling

Szarad:
I just noticed that I had a plugin turned on, that reformatted the whole file. If it is too difficult to review with all these stylistic changes, I can commit again. Sorry for that.

Pipfile
```diff
@@ -15,6 +15,10 @@ aiofiles = "*"
 babel = "*"
 aiohttp = "*"
 sentry-sdk = "*"
+platformshconfig = "*"
+redis = "*"
+fakeredis = "*"
```
lsmith77: I guess these two dependencies should be moved to the dev-packages section

app/main.py
```diff
@@ -46,6 +49,16 @@
 from pydantic import BaseSettings
 from app.categories import categories
 
+
+from redis import Redis
+#from platformshconfig import config
```
lsmith77: I guess this can be removed

app/main.py
```diff
@@ -179,14 +197,25 @@ async def custom_http_exception_handler(request, e):
               "Kollektivgesellschaft", "Team", "Organization", "Gliederung"]
 gender_false_positive = genderdenom_false_positives["False_positives"].tolist()
 
-#corporate false positive
-#DB
-corporate_false_positive = ["stark", "starke", "starkes", "starker", "Führungskraft", "Führungskräfte", "Führungskräften"]
+# corporate false positive
+# DB
+corporate_false_positive = ["stark", "starke", "starkes",
```
lsmith77: this can be removed, ie. we don't need the hardcoded false positive list anymore

Pipfile
Szarad: good point, yes fakeredis and pytest-asyncio  is need for testing, I'll move it

app/main.py
Szarad: What exactly do you mean?
I can remove import json  and JSONResponse as it is not used. I can also remove commented imports.

app/main.py
Szarad: I think I don't understand. 
I thought these are the only options:

    store_context: Optional[bool] = True
    primary_language: Optional[str] = "de-DE"
    preferred_languages: Optional[str] = "de,en"
    preferred_variants: Optional[str] = "de-DE,en-GB"
    german_gender_ending: Optional[str] = ":in"
   _gendereddenom_ending = {"/in": "/in", "/-in": "/-in", "_in": "_in", "*in": "\\*in", ":in": ":in", "In": r"In\b"}
   disabled_categories: Optional[List] = ""
   gendered_roles_format: Optional[GenderedRolesFormatType] = "inclusive_gender"

Where user can store false positive ?

app/main.py
lsmith77: just that we don't need to keep commented out code

app/main.py
lsmith77:
```
corporate_false_positive = ["stark", "starke", "starkes", "starker", "Führungskraft", "Führungskräfte", "Führungskräften"]
```

These are "false positives" we hardcoded for DB. this list needs to be customizable by corporation in the dashboard and then synced via your API into Redis

Szarad: ah ok it is just for Redis? I will then add another class based on Config with attribute corporate_false_positive, so when dashboard triggers POST on redis it know what to expect.

lsmith77: yes .. we will not allow this config via the browser extension, ie. it cannot be submitted via a /check request, it has to come from Redis (we use an empty `[]` as default)

lsmith77: moved to https://github.com/witty-works/nlp_api/issues/198

Issue #162, opened 2021-10-28
Update logging


lsmith77:
<https://docs.google.com/document/d/1_04xIRx9esZ8JfvPvWd6N2SghWH0cC2KisbGcU0WUdg/edit?disco=AAAAQZZQRYo&usp_dm=true>

<p></p>

- Replace Sentry

- Implement the new logging features mentioned in the privacy policy

  - 100 character before/after by sending this back as the "context"

- Log client/browser infos

lsmith77:
Link to feature: https://wittyworks.productboard.com/feature-board/planning/features/10452987

PR #177, opened 2021-11-10
in place of storing the text, store relevant metrics in the logs

lsmith77:
#162 forgot to make this change. the idea is that we only log some data points about the data but not the actual data with each request.

note that for events like a click on an alternative or ignore we store the 100 chars before/after but this requires notification from the browser and handled via the "event" case where we store everything. however when storing the response we hide the context giving users some control over where we store context.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -24,7 +24,7 @@ import copy
 from typing import Optional
 from functools import lru_cache
 
-from fastapi import FastAPI, Request, HTTPException, BackgroundTasks, Depends, status
+from fastapi import FastAPI, Request, Header, HTTPException, BackgroundTasks, Depends, status
 
 from fastapi.openapi.docs import get_swagger_ui_html
 from fastapi.openapi.utils import get_openapi
@@ -273,7 +273,6 @@ def get_current_username(credentials: Optional[HTTPBasicCredentials] = Depends(s
 def get_root():
     return RedirectResponse(url='/form', status_code=301)
 
-
 @app.get("/lt")
 def get_root():
     return languagetool_url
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -404,16 +404,31 @@ def language_rules(user_request_in: RequestIn, lang: Lang):
 
     return list_results
 
+def serialize_request_data(user_request_in: RequestIn):
+    data = {}
+    data["text"] = {
```
lsmith77, line 413: @lnazarenko the counting of the words and sentences is quite naiv here .. should we use some better approach here using NLP?

lnazarenko: @lsmith77 Do you mean to write a couple NLP analysis functions to get insights from logs data?

lnazarenko: "length": len(user_request_in.text),
         "words": len(user_request_in.text.split(" ")),
         "sentences": len(user_request_in.text.split(".")),
     }
you already calculated here the length of the whole text, words and sentences

lnazarenko: In general, yes, we can add NLP approach here. We need to decide what we want to extract. Probably not just the length of the words

lsmith77: yes .. if its important to more accurately compute number of words and sentences than just counting spaces and dots. since of course multiple spaces would lead to a wrong word count and of course also sentences end with ? and ! and people might type ".."

lnazarenko (approved):

app/main.py
lnazarenko: Ok, I can add this, no problem. But why do we need to calculate the exact number of words or sentences in the logs? We wrote that we only collect 100 symbols around the highlighted word.

lsmith77: so the point is that we do not want to store the entire text so we store what we can without causing privacy issues. the reason I asked is because I wanted to get your opinion what data would actually be useful for us to collect and how accurate it needs to be.

as for the 100 chars before and after, as per our current privacy policy we only collect them when an alternative is either click on or if a highlighted words is being ignored.

lsmith77: anyway .. this looks like a topic we should briefly discuss tomorrow afternoon in person.

app/main.py
```diff
@@ -404,16 +405,35 @@ def language_rules(user_request_in: RequestIn, lang: Lang):
 
     return list_results
 
+def serialize_event_data(user_request_in: RequestIn):
+    data = user_request_in.dict()
+
+    return jsonable_encoder(data)
+
+def serialize_request_data(user_request_in: RequestIn, user_agent: str = Header(None), origin: str = Header(None)):
```
lsmith77: https://fastapi.tiangolo.com/tutorial/header-params/

app/main.py
lsmith77: decision was made to only store the text length

PR #186, opened 2021-11-16
upgrade to python 3.9

lsmith77:


lsmith77:
I know there is already 3.10 but I figured we make a small step first and 3.10 was only released very recently.

Szarad (approved):
looks good, shouldn't change code behavior between this version

PR #210, opened 2021-11-23
Fix problem with Docker tag.

Szarad:


.github/workflows/deployAzure.yaml
```diff
@@ -17,15 +17,21 @@ jobs:
           docker build . -t ${{ secrets.REGISTRY_LOGIN_SERVER }}/nlpapi:${{ github.sha }}
           docker push ${{ secrets.REGISTRY_LOGIN_SERVER }}/nlpapi:${{ github.sha }}
           # release version
-          docker tag ${{ secrets.REGISTRY_LOGIN_SERVER }}/nlpapi:${{ github.sha }} ${{ secrets.REGISTRY_LOGIN_SERVER }}/nlpapi:${{ github.ref_name }}
-          docker push ${{ secrets.REGISTRY_LOGIN_SERVER }}/nlpapi:${{ github.ref_name }}
+          ref=${{ github.ref_name }}
+          tag=${ref////-}
+          echo $tag
```
lsmith77, line 22: I guess you intentionally left the echo's in ?

Szarad: yes, I think it would be good to see a tag during build.

PR #212, opened 2021-11-24
by default do not read rules from redis

lsmith77:
since we currently do not yet have any corporate rules in redis, lets disable reading from redis for now.


Szarad (approved):
good idea to add it, thank you

PR #216, opened 2021-11-26
Add fasttext to LanguageTool image.

Szarad:
Install and use fasttext library in LanguageTool image. It was tested with examples from issue https://github.com/witty-works/nlp_api/issues/126


lsmith77:
FYI I have upsized our plan from 5GB to 10GB. is that enough? you will need to then adjust the disksize in the .platform.app.yaml file to leverage the new space .. ie. add 5GB there.

https://docs.platform.sh/configuration/app/storage.html#disk

lsmith77:
FYI we have an issue with the platform.sh build cache .. I ran `platform project:clear-build-cache` and then `platform e:redeploy` and it managed to deploy your PR on platform.sh after all

Szarad:
Ok, thank you for information. 5GB should be more then enough, uncompressed fastext model has 126MB file size and I used the compressed one (917kB). As it managed to deploy afterwards, should I add new space anyway?

Szarad:
In Spacy Test changes are working: https://main-bvxea6i-qh3uq7skrqxzg.de-2.platformsh.site/docs
I would try to merge and see how to behave in dev, if needed we can revert it.

PR #226, opened 2021-12-01
output cache dir when copying LT/FastText

lsmith77:


languagetool/.platform.app.yaml
```diff
@@ -60,13 +61,20 @@ hooks:
     else
       echo "Using cached archive of ${FT} ${FT_VERSION} ..."
     fi
-    echo "Installing archive of ${FT} ${FT_VERSION} ..."
-    cp -R "${PLATFORM_CACHE_DIR}/${FT_DIR}"/* ./"${FT}"
+
+    echo "Installing archive of ${FT} ${FT_VERSION} from '${PLATFORM_CACHE_DIR}/${FT_DIR}' ..."
+    mkdir -p ${FT}
```
lsmith77, line 65: @Szarad this mkdir was missing

languagetool/.platform.app.yaml
```diff
@@ -60,13 +61,20 @@ hooks:
...
+    cp -R "${PLATFORM_CACHE_DIR}/${FT_DIR}"/* "./${FT}"
 
     export FT_MODEL="lid.176.bin"
     if [ ! -f "$FT_MODEL" ]; then
-      echo "Download ${FT_MODEL} ..."
+      echo "Downloading ${FT_MODEL} ..."
```
lsmith77, line 70: added caching here as well

languagetool/.platform.app.yaml
Szarad: mkdir in line 58 wasn't enough?

languagetool/.platform.app.yaml
Szarad: they don't do versioning of this file, but maybe it safer to have it like that as well

lsmith77: exactly that was my thinking

languagetool/.platform.app.yaml
lsmith77: ah didn't see that one .. the issue is that this was only called while caching but if the cache exists, that mkdir was never reached again

lsmith77: FYI just pushed to remove the mkdir on line 58

Szarad: ah I understand now, thanks for the fix

Szarad (approved):
thanks for changes!

PR #233, opened 2021-12-06
rules for new category openly discriminating English

lnazarenko:


lsmith77:
@lnazarenko can you reimport the training data as some of the subcategory names changed

PR #234, opened 2021-12-06
posthog analytics

lsmith77:
For now this just makes it possible to configure which IDs should have their analytics redirected to posthog. 

This way we can start collecting data from Witties so that we can start playing with posthog to build out statistics.


Szarad (approved):
Looks good for me

PR #242, opened 2021-12-07
Features/improve tests

Szarad:


Szarad:
Test should be run as usual using: `$pytest -vv`
Updates should be done: `$pytest --snapshot-update` (only run when changing input for test!)

Pipfile
```diff
@@ -19,6 +19,7 @@ opencensus-ext-azure = "*"
 redis = "*"
 platformshconfig = "*"
 fakeredis = "*"
+pytest-snapshot = "*"
```
lsmith77: this should probably be in the dev section?

lsmith77 (changes requested):
amazing!!! this will save soooo much time.

can you look at moving the dep to --dev?

and you can merge

Pipfile
Szarad: ah thank you, I will change it.

lsmith77 (approved):

PR #249, opened 2021-12-08
Add alternatives to spellchecker ignore file.

### LINKED ISSUE #237: Kollegschaft is not recognized by Language Tool
## Is situation

<p></p>

- On <https://default.api.witty.works/form>, I write "liebe Kollegen".

- I choose as alternative "Kollegschaft"

- It get's highlighted, since the word is not recognized in the language tool (see attached video)

<p></p>

## ACC. Criteria

The word "Kollegschaft" is recognized by our algorithm

Szarad:
fixes https://github.com/witty-works/nlp_api/issues/237\

Kund:in/Kunde

=>

Kundin
Kunde
Kund:in
Kund*in
Kund/in
Kund/-in
KundIn

Kund:innen und Kunden

Kundinnen
Kunden
und
Kund:innen
Kund*innen
Kund/innen
Kund/-innen
KundInnen


Szarad:
With encapsulation first approach (ignoring words using ignore.txt) is a way to go and worked well tested locally with Docker containers.

languagetool/ignore.txt
```diff
@@ -0,0 +1,2850 @@
+(gestellte) Anträge
+(von Berufs wegen) beratende Personen
```
lsmith77: I think we will need to split (on ` ` and `/`) each of the alternatives into words and strip parenthesis

languagetool/ignore.txt
```diff
@@ -0,0 +1,2850 @@
...
+Abdichter\*in
+Abdichter\*innen
+Abdichter\/-in
+Abdichter\/-innen
+Abdichter\/in
+Abdichter\/innen
+Abdichter\:in
```
lsmith77: if this works we might be able to revert https://github.com/witty-works/nlp_api/commit/7ad5648fd3463d5e8602157f5d9f7a1985bbe050

add_alternatives_to_ignore.py
```diff
@@ -0,0 +1,71 @@
+import csv
+import json
+from collections import defaultdict
+import string
+
+columns = defaultdict(list) 
+all_words = []
+
+with open('training_data/GenderedNoun_DE.csv') as f:
```
lsmith77: we will eventually need to run this across all training_data files.

I think the script should also call https://lt.default.api.witty.works/v2/check first to verify if the is actually seen as a spelling issue by LT so that our list doesn't grow too long.

lsmith77 (commented):
Nice, I guess with using `\` it is now possible to also ignore :, * etc?

add_alternatives_to_ignore.py
Szarad: Should I done it for this PR or in next iteration?

lsmith77: lets do this in this iteration. I am party concerned about the performance impact.

languagetool/ignore.txt
Szarad: hmm I can do that, but I put the form as it is in PL_all_clean, e.g. 
`['Mittler~innen~ und ~Mittler', 'Vermittlung',  'Personen, die helfend vermitteln', 'Personen, die vermittelnd (zwischen  Personen, Parteien) auftreten']`

lsmith77: the goal here is to ignore words that LT would mark as a spelling error.

so in the above example the list of candidates words are:
* Mittler
* Mittler*innen
* Mittler:innen
* Mittler/innen
* Mittler/-innen
* MittlerInnen
* und
* Vermittlung
* Personen
* die
* helfend
* vermitteln
* vermittelnd
* zwischen
* Parteien
* auftreten

Each word would be send to LT as the `text` param with the `language` as per the ending of the training data file

If LT gives any results, we add it to the ignores.txt file

add_alternatives_to_ignore.py
Szarad: Check was added (line 72-77 in new add_alternatives_to_ignore.py).  In the end for German it added 450 word... Still a lot, but I guess we don't have much choice?

lsmith77: yeah .. I guess we will also ask Tracey to go over the list to make sure there are no typos in there

add_alternatives_to_ignore.py
```diff
@@ -0,0 +1,78 @@
...
+from collections import defaultdict
+import string
+import requests
+
+test_text = "Liebe "
+url_test = "https://lt.default.api.witty.works/v2/check"
+columns = defaultdict(list)
+all_words = []
+
+with open("training_data/GenderedNoun_DE.csv") as f:
```
lsmith77: FYI I have renamed this file to `gendered_noun_DE.csv`

Szarad:
450 words in the end, I think the ignore.txt file looks correct now, beside 
1. `Matrosinnen/`, which looks like this in the csv file: `'Matrosinnen~ und ~Matrosen'`
In the middle step (before calling /check endpoint) I found problems with following words:
2. `Nutzerinnen/`, which looks like this in csv: `'Nutzerinnen~ und ~Nutzer'`
3. `'Kauffrau//`, which looks like this: `'Kauffrau~/~Kaufmann'`

add_alternatives_to_ignore.py
```diff
@@ -0,0 +1,134 @@
...
+with open("languagetool/ignore.txt", "r") as readfile:
+    used_words = [line.strip() for line in readfile]
+
+with open("languagetool/ignore.txt", "w") as myfile:
+    for word in words_post:
+
+        if word not in used_words:
+            response = requests.post(
+                url_test,
+                data={"text": word, "language": "auto"},
```
Szarad: previously: `   data={"text": (test_text + word), "language": "auto"}` (I get rid off test_text)

add_alternatives_to_ignore.py
```diff
@@ -0,0 +1,134 @@
...
+            )
+            if response.status_code == 503:
+                continue
+            if response.json()["matches"]:
```
Szarad: add this if:
`  if ( response.json()["matches"][0]["message"]  == "Möglicher Tippfehler gefunden."):`

to double check that error is misspelling

add_alternatives_to_ignore.py
```diff
@@ -0,0 +1,134 @@
+import csv
+import json
+from collections import defaultdict
+import string
+import requests
+
+url_test = "https://lt.default.api.witty.works/v2/check"
```
Szarad: Get rid off `test_text = "Liebe "`

PR #250, opened 2021-12-08
new English style category

lnazarenko:
English style category. It has only one third of the filled alternatives. As soon as Tracey will finish it will do the rules update.


lsmith77:
FYI the platform.sh failure is unrelated.

PR #261, opened 2021-12-15
re-add sentry

lsmith77:
We had Sentry before which I removed because I didn't want to have them in our processors list as their hosting is in the US however we will eventually self-host

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -7,6 +7,7 @@ import json
 import secrets
 import aiohttp
 import copy
+import sys
 
 from fastapi import (
     FastAPI,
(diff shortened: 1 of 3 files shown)


app/sentry.py
```diff
@@ -0,0 +1,27 @@
+import sentry_sdk
+from sentry_sdk.integrations.aiohttp import AioHttpIntegration
+from posthog.sentry.posthog_integration import PostHogIntegration
+
+# Sentry SDK set up with PostHog integration
+def set_up_sentry_sdk(version, settings):
+    if not settings.sentry_dsn or settings.testing == True:
+        return False
```
Szarad: this line is causing the problem, sentry_sdk is set to False during testing

Szarad (changes requested):

app/sentry.py
Szarad: We can here `return None` and check the value in main.py in line 439

app/main.py
```diff
@@ -409,6 +433,11 @@ def log(
 async def check_query(
     request: Request, user_request_in: RequestIn, background_tasks: BackgroundTasks
 ):
+    if not user_request_in.id == None:
+        posthog.identify(id)
+
+    sentry_sdk.set_context("request", clean_requestin_data(user_request_in))
```
Szarad:
```python
if sentry_sdk: 
    sentry_sdk.set_context("request", clean_requestin_data(user_request_in))
 ```

**Review by Szarad — CHANGES_REQUESTED** [review]

File: app/sentry.py [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+from posthog.sentry.posthog_integration import PostHogIntegration
+
+# Sentry SDK set up with PostHog integration
+def set_up_sentry_sdk(version, settings):
+    if not settings.sentry_dsn or settings.testing == True:
+        return None
+
+    integrations = [AioHttpIntegration()]
+
+    if settings.training_data_enabled:
```
lsmith77: note right now I have not been able to get the posthog integration to work :-/

lsmith77: see also https://github.com/PostHog/posthog-python/issues/46

Szarad (approved):
Looks good to me

PR #269, opened 2021-12-20
determine category from sub-category

lsmith77:


app/main.py
```diff
@@ -947,8 +952,8 @@ def IsItFalsePositive(word, false_positive):
 
 
 def GenderedDenomEnd(config: Config, lang, full_text):
-    category = "gendered"
     subcategory = "gendered_denominations_ending"
+    category = categories[subcategory]["category"]
```
lsmith77, line 956: with this line we automatically determine the correct category for a subcategory, which means it will get automatically updated if the notion page is re-imported.

PR #271, opened 2021-12-21
german unc_bias exept agentic

lnazarenko:


lnazarenko:
- German Uncon Bias category except agentic. Now there are two functions: one handle triggers nouns to show singular and plural forms of alternatives, mostly consist of exaggerated subcategory). Second function handle not noun triggers (adj and verbs) and sentences (consist exaggerated, previous un.bias). Please note, previously un. bias didn't show alternatives
- Small fix for open disc with GetNonNounLowerCased
- fix test for unc bias word Ninja

PR #284, opened 2022-01-05
Rules refactoring for german language.

Szarad:


lsmith77:
looks already a ton better!

note I think we should however pass the rules data to the functions explicitly.

1) for test-ability
2) since some of the functions are generic

also we probably want to eventually convert the rules from variables to an object. but this can be done once we have completed the refactorings you started here.

PR #291, opened 2022-01-12
Fix wrong references.

Szarad:


lsmith77:
thx .. its a bit concerning that our test suite has not shown these issues to us. how did you find these issues?

Szarad:
It is concerning, I can add test, but we need some more general overview on use cases to test. I discovered it, rebasing another branch...

lsmith77:
ok. the rules circle is working on expanding our example sentences. once that is done we update our tests to hopefully increase coverage. 

speaking of code coverage, we actually already generate coverage reports via github actions. we just don’t visualize them anywhere. so I will create a task to visualize code coverage of our tests

PR #293, opened 2022-01-14
Blackfire integration - remove unnecessary configs

sumerc:


Szarad (approved):
ok, let's give it a try

PR #301, opened 2022-01-17
improve exception handling to clean up sentry issue list

lsmith77:
fixes NLP_API-2F

* do not throw an exception if lang could not be determined
* raise exception of LT returns a 500

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -12,6 +12,7 @@ import sys
 from fastapi import (
     FastAPI,
     Request,
+    Response,
     HTTPException,
     BackgroundTasks,
     Depends,
@@ -29,7 +30,7 @@ from fastapi.exception_handlers import http_exception_handler
 from starlette.responses import RedirectResponse, PlainTextResponse
 from starlette.exceptions import HTTPException as StarletteHTTPException
 
-from typing import Optional
+from typing import Optional, Union
 
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from sentry_sdk import configure_scope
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -216,20 +218,15 @@ def get_root():
     return RedirectResponse(url="/form", status_code=301)
 
 
-@app.get("/exception", response_model=ResultsOut)
+@app.post("/exception")
```
lsmith77, line 221: this end-point is just for testing and I adjusted to mimic more closely `/check`

app/models.py
```diff
@@ -313,12 +313,27 @@ def getGenderedRolesFormatInclusive(alternative, german_gender_ending):
...
 class ResultsOut(BaseModel):
     results: List[ResultOut]
     language: str
 
     def factory(results, lang):
-        return ResultsOut(results, lang.lang)
+        if lang != None:
+            lang = lang.lang
+
+        return ResultsOut(results, lang)
```
Szarad, line 336: Should we call ResultsOut when lang is None?

app/models.py
```diff
@@ -313,12 +313,27 @@ def getGenderedRolesFormatInclusive(alternative, german_gender_ending):
...
+        object.__setattr__(self, "details", details)
+
+
 class ResultsOut(BaseModel):
     results: List[ResultOut]
     language: str
 
     def factory(results, lang):
-        return ResultsOut(results, lang.lang)
+        if lang != None:
```
lsmith77, line 333: this is technically not needed but just to be sure

Szarad:
ok, then looks good to me

Szarad (approved):

app/models.py
lsmith77: no .. technically this should never happen

PR #305, opened 2022-01-17
Add check for token morph number.

### LINKED ISSUE #302: check if morph.get("Number") returns a list before accessing [0]
follow up for https://github.com/witty-works/nlp_api/issues/296

Szarad:
fixes #302

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -340,6 +340,17 @@ async def get_redis(user: str):
 
 
 # Functions
+
+
+def is_number_list_empty(number, token):
+    if not number:
+        logging.error(
+            "List of token morph number: %s for token/word: %s", number, token
+        )
+        return True
+    return False
+
+
 def set_browser_id(request: Request, id):
     request.state.browser_id = str(id)
 
@@ -999,7 +1010,10 @@ def GenderedDenomAnalysisDE(
                 subcategory,
             ) in gender_words_alternatives:
                 if tokens[i].lemma_ == word:
-                    if tokens[i].morph.get("Number")[0] == "Sing":
+                    token_morph_number = tokens[i].morph.get("Number")
+                    if is_number_list_empty(token_morph_number, tokens[i]):
+                        continue
+                    if token_morph_number[0] == "Sing":
                         list_tokens.append(
                             ResultOut.factory(
                                 config,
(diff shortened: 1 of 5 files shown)


app/main.py
```diff
@@ -999,6 +1012,8 @@ def GenderedDenomAnalysisDE(
                 subcategory,
             ) in gender_words_alternatives:
                 if tokens[i].lemma_ == word:
+                    if not isNotListFromToken(tokens[i]):
```
lsmith77: I thought the issue is caused by accessing `morph.get(“Number)[0]`

Szarad: yes, it is

Szarad: What's wrong in this line? If morph.get("Number") is empty, then the code should not go further, so continue statement is called

Szarad: empty `morph.get("Number")` list -> function `isNotListFromToken(token)` return False -> go inside `if not` statement -> triggers `continue`

lsmith77: ah .. sorry misread the code. I guess now we are calling `get(“Number”)` twice. no clue how expensive it is.

app/main.py
```diff
@@ -340,6 +340,19 @@ async def get_redis(user: str):
 
 
 # Functions
+
+
+def isNotListFromToken(token):
+    number = token.morph.get("Number")
+    if len(number) > 0:
+        return True
+    else:
```
lsmith77: the `else` isn’ needed here since the `if` would `return` already

Szarad: It can be rewrite as:
```
def isNotListFromToken(token):
    number = token.morph.get("Number")
    if len(number) > 0:
        return True
    
    logging.error(
            "List of token morph number: %s for token/word: %s", number, token
        )
    return False`
```
Is that what you mean?

app/main.py
Szarad: true, better change it

app/main.py
```diff
@@ -340,6 +340,17 @@ async def get_redis(user: str):
 
 
 # Functions
+
+
+def isNotListFromToken(token):
+    number = token.morph.get("Number")
+    if len(number) == 0:
```
Szarad: Logic changed to not call `token.morph.get("Number") ` too many time

Szarad:
> 



> one last bit, the name of the function should not likely be changed

What do you mean?

lsmith77:
`isNotListFromToken` does not represent well what this function does .. 

ie. it now gets the number and logs if its not a list.

app/main.py
```diff
@@ -340,6 +340,17 @@ async def get_redis(user: str):
 
 
 # Functions
+
+
+def is_number_list_empty(number, token):
```
Szarad, line 345: I rewrite the function to be more pythonic. Function now return True if number list is empty.

lsmith77 (approved):

PR #315, opened 2022-01-18
286 generate false positive

### LINKED ISSUE #286: generate titles false positive file
partially revert <https://github.com/witty-works/nlp_api/commit/301d73f9f5b380458def9a49ce0abd6cf49ce955> to handle cases like "Fahrzeugtechnikerinnen und Fahrzeugtechniker" and "Fahrzeugtechnikerinnen/Fahrzeugtechniker"

<p></p>

probably best to integrate this into add_alternatives_to_ignore.py

<p></p>

<p></p>

Szarad:
fixes #286 

False positive list is in training_data/de-DE/titles_false_positives.csv


lsmith77 (commented):
I have not reviewed the code in detail yet. but we also need to partially revert https://github.com/witty-works/nlp_api/commit/301d73f9f5b380458def9a49ce0abd6cf49ce955 so that these false positives are taken into account again.

Szarad:
@lnazarenko these are the columns and csv files that I am checking. Let me know if this is correct:
```
rules["training_data/de-DE/agentic.csv"] = ["Alt_split"]
rules["training_data/de-DE/gendered_noun.csv"] = ["Sg_all_clean", "Pl_all_clean"]
# following files do not add anything to false_positive list
rules["training_data/de-DE/hollow_sentences.csv"] = [
    "Alternative_Plural_split",
    "Alternative_Singular_split",
]
rules["training_data/de-DE/hollow_words.csv"] = [
    "Alternative_Plural_split",
    "Alternative_Singular_split",
]
rules["training_data/de-DE/open_dis_words.csv"] = ["Alt_split"]
rules["training_data/de-DE/style_sentences.csv"] = ["Alt_split"]
rules["training_data/de-DE/style_words.csv"] = ["Alt_split"]
rules["training_data/de-DE/ub_no_noun_words.csv"] = ["Alt_split"]
rules["training_data/de-DE/ub_noun_words.csv"] = ["Alt_split"]
rules["training_data/de-DE/ub_sentences.csv"] = ["Alt_split"]
```
and for these files I am extracting sentences like "Word(f) und Word(m)" and "Word(f)/Word(m)" (some of the files above didn't add anything)
then I am removing the ones that are not highlighted

lnazarenko (approved):
@Szarad From my side looks good now. Thank you for the short call!  Just to summarise: 1. To make the generate_false_positive more generic I will add Alt_split column to gendered_noun German that also would be a small step forward to make automated update of the rules. 2. Just to keep in mind when you manipulate with the csv data. CSV pre-built python module is nice and fast enough if the dataset consist less then 1000 rows. For the larger datasets it is better to use pandas library to work with csv files. But here we don't need to care about the speed to much since this a preprocessing file it doesn't influent to the NLP API performance. Just for your curiosity :)  Here is performance comparison https://stackoverflow.com/questions/62139040/python-csv-module-vs-pandas

Szarad:
@lsmith77 should we incorporate that  or work first on make it more generic? "Our" fasle_positive list is read here, that is right?
```
false_positive = get_false_positive(
    rules["de-DE"]["false_positive_agentic_const"],
)
```
So I need to change name of the csv file or name here

Szarad:
Ok, I changed the script to be more generic. It is in the branch `286-generate-false-positive-generic`
[https://github.com/witty-works/nlp_api/blob/286-generate-false-positive-generic/generate_false_positive.py](url)
@lnazarenko now I need to have Alt_split column in gendered_noun.csv  and I think in hollow_sentences.csv

Szarad:
ok from my point of view all done. I also tested it manually. 
result on dev branch (e.g. `Abdichterinnen und Abdichter`):
```
{
  "results": [
    {
      "text": "Abdichter",
      "context": "Abdichterinnen und Abdichter",
      "category": "gendered",
      "subcategory": "titles",
      "start": 19,
      "end": 28,
      "alternatives": [
        "Abdichtung",
        "Abdichter:in",
        "Abdichterin/Abdichter"
      ],
      "label": "Geschlechtsspezifisch: Titel",
      "reason": "Das männliche Generikum spricht nicht alle Geschlechter oder Geschlechtsidentitäten an. Viele Menschen fühlen sich daher nicht in den Dialog einbezogen.",
      "solution": "Verwenden Sie eine Schreibweise, die das weibliche Geschlecht sowie auch andere Geschlechteridentitäten, die nicht einem binären Verständnis von Geschlecht folgen, anspricht."
    }
  ],
  "language": "de"
}

```
result on this branch:
```
{
  "results": [],
  "language": "de"
}
```
Unless @lnazarenko  @lsmith77 you want to change some name or sth else, it could be integrated

lsmith77:
@Szarad will look over the PR but FYI there are conflicts that need to be resolved

Szarad:
rebasing now

lsmith77 (commented):
can you add something to the README?

generate_false_positive.py
```diff
@@ -0,0 +1,139 @@
...
+)
+import re
+import argparse
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-u",
+    "--URL",
+    help="Api url which should call check on training data, e.g. http://localhost:8000/check",
```
lsmith77: maybe localhost should be the default .. 

also super small thing but likely it would be `http://localhost:8000/v2/check` and not `http://localhost:8000/check`

lsmith77: ah nevermind .. I guess this is not the URL to an LT server .. but to our server .. right?

in that spirit I guess on that server the `training_data/de-DE/gender_false_positive.csv` needs to be emptied before running, right?

Szarad: To our server, right. But I can add default localhost. The write function overwrite the file, so it doesn't have to be empty beforehand.

lsmith77: about overwriting the file ..

my point is if the file has entries in it, then those will not be seen as false positives when running the script and therefore not re-added to the file. that being said .. I didn't run the script yet to test this.

Szarad: Turns out I would need to add some validation in rules.py for that (to not have errors like `pandas.errors.EmptyDataError: No columns to parse from file`). I am doing it

lsmith77: hmm .. a quick fix could be to just add a dummy entry

Szarad: ok, I tried to add check as `if os.path.isfile(csv_path) and os.path.getsize(csv_path) > 0:` , but I got errors like that `ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().` Anyway I guess for that we need to open another tickets, because whole structure of rules would need to be changed. And we need to add this, so if any of csv file will be deleted, our app won't crash. I can open the ticket for that.

 Coming back to this PR, how do you want to add dummy entry? Should I add another script, which just clean csv file and add dummy entry or just add instructions to read.me? Then if  we agree on some dummy entry like {"test":"test" } I can check in generate_false_positive if csv contains this dummy entry. And later when we refactor rules I can check if csv file is empty or it doesn't exist. 
 
 I hope you can understand what I mean.

lsmith77: I think just documenting it as manual steps in the README is fine

Szarad: * dummy data must be `"False_positives"`, I will soon add changes. It should make sens then

README.md
```diff
@@ -206,7 +206,29 @@ pipenv run python -m analyze_rules -i [CSV export]]
 ```
 pipenv run python -m add_alternatives_to_ignore
 ```
+## Update the false positive list 
+1. Remove gender_false_positive.csv file (keep the old copy for safety):
```
lsmith77: since we have the file in Git, I don't think this step (and step 5 is really needed.

generate_false_positive.py
```diff
@@ -0,0 +1,191 @@
...
+            potential_false_positive[locale].append(w)
+
+
+def checked_false_positive_list(api_url, potential_false_positive):
+    false_positive_checked = []
+    for locale in potential_false_positive:
+        for word in potential_false_positive[locale]:
+            response = requests.post(
+                api_url,
+                json={"text": word, "lang": "auto"},
```
lsmith77: please change `auto` to the specific locale and also set `disabled_categories` to `orthography`

lsmith77:
ideally we first merge https://github.com/witty-works/nlp_api/pull/316 and https://github.com/witty-works/nlp_api/pull/323

then rebase this PR

PR #316, opened 2022-01-18
move locale prediction to python

### LINKED ISSUE #308: move language detection from LT to python API

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/308

several tests had to be updated since we now prefer `en-US` over `en-GB`


.github/workflows/main.yaml
```diff
@@ -30,11 +35,32 @@ jobs:
           pipenv run python -m spacy download en_core_web_sm
           pipenv run python -m spacy download de_core_news_sm
 
-      - id: cache-po-files
-        uses: actions/cache@v1
```
lsmith77, line 34: I switched over to `cache@v2` since only `v2` is able to cache individual files https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows

README.md
```diff
@@ -39,6 +39,7 @@ docker run  -p 8000:8000 --name nlp --network "bridge" --env languagetool_api=ht
 pipenv install --dev
 pipenv run python3.9 -m spacy download en_core_web_sm --no-cache-dir
 pipenv run python3.9 -m spacy download de_core_news_sm --no-cache-dir
+wget -P training_data https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin
```
lsmith77, line 42: this will also need to be added to the Docker setup

app/models.py
```diff
@@ -72,19 +88,47 @@ class Config(BaseModel):
...
+                if lang not in Config._supported_langs:
+                    raise ValueError("Contains not supported preferred_languages: " + v)
+
+            return langs
+
+        return []
+
+    @validator("preferred_variants", pre=True)
+    def valid_preferred_variants(cls, v):
+        if isinstance(v, str) and v != "":
```
Szarad, line 122: validator is not working after changing type of class attribute (from str to List)
can be rewrite to:
```
if isinstance(v, list) and v != []:
    for lang in v:
        if lang not in Config._supported_locales:
            raise ValueError("Contains not supported preferred_variants: %s" % lang)
    return v
```

app/models.py
```diff
@@ -93,26 +137,48 @@ def split_string_values(cls, v):
...
 
+    @validator("primary_language", pre=True)
+    def valid_primary_language(cls, v):
+        if v not in Config._supported_locales:
+            raise ValueError("Not supported primary_language: " + v)
+        return v
+
+    @validator("preferred_languages", pre=True)
+    def valid_preferred_languages(cls, v):
+        if isinstance(v, str) and v != "":
```
Szarad, line 164: validator not working, `v` is a List now
Can be rewrite to:
```
if isinstance(v, list) and v != []:
    for lang in v:
        if lang not in Config._supported_langs:
            raise ValueError("Contains not supported preferred_languages: %s" % lang)
    return v
```

app/models.py
```diff
@@ -93,26 +137,48 @@ def split_string_values(cls, v):
...
+                if lang not in Config._supported_langs:
+                    raise ValueError("Contains not supported preferred_languages: " + v)
+
+            return langs
+
+        return []
+
+    @validator("preferred_variants", pre=True)
+    def valid_preferred_variants(cls, v):
+        if isinstance(v, str) and v != "":
```
Szarad, line 178: validator not working, the same as above

app/models.py
```diff
@@ -72,19 +88,47 @@ class Config(BaseModel):
...
+
+    @validator("primary_language", pre=True)
+    def valid_primary_language(cls, v):
+        if v not in Config._supported_locales:
+            raise ValueError("Not supported primary_language: " + v)
         return v
 
+    @validator("preferred_languages", pre=True)
+    def valid_preferred_languages(cls, v):
+        if isinstance(v, str) and v != "":
```
Szarad, line 108: Validator is not working, because `v` is a List now, the same as above

Szarad (changes requested):

app/models.py
lsmith77: I fixed it so that it now supports both strings (comma separated) and lists

Szarad (approved):
Tested, looks good from my side.

PR #318, opened 2022-01-19
revert c576dae2aa22da136a7670824f1b580d99b5e380 since we now have the custom false positive file

lsmith77:


Szarad (commented):
I talked with Elena regarding PR 315 to have the make script more generic, for that we need to have the same column across all csv files. We will open follow up ticket, but we agree for now we can incorporate 315 PR. What do you think?

lsmith77:
this is related to LT where we already generate the false positive file, ie. ignore.txt https://github.com/witty-works/nlp_api/blob/dev/add_alternatives_to_ignore.py#L86

Szarad:
ah ok

Szarad (approved):
approving then, sorry for the confusion

PR #319, opened 2022-01-19
fix setting the sentry user to the browser id

lsmith77:
essentially drop the use of custom exception handler and also drop the posthog integration

I tried to get it to work with the custom handler but ended up stumbling on a lot of github issues. in the end lets just simplify things.

ideally at some point posthog also gets their sentry integration to work right but for now I think its not so critical to know inside posthog which user was affected until we get the login functionality.


Szarad (approved):
That's a pity, but checked. Looks fine, just removing handler & posthog

PR #321, opened 2022-01-19
test cleanup

lsmith77:


tests/test_api.py
```diff
@@ -12,6 +11,10 @@
 client = TestClient(app)
 
 
+def get_dirs(path):
+    return list(file for file in Path(path).iterdir() if not file.name.startswith("."))
```
Szarad: very small thing, name as `directory` or `path` rather than `file` would be more readable for me

Szarad (approved):
I added just a small comment, beside that looks good.

PR #322, opened 2022-01-19
ignore @ and # sign

### LINKED ISSUE #320: hashtags
<https://forum.languagetool.org/t/ignore-hashtags-user-mentions/3305/14>

lsmith77:
fixes #320

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -392,6 +392,7 @@ async def languagetool_rules(user_request_in: RequestIn):
     list_results = []
     lang = None
     config = user_request_in.config
+    ignore = ["@", "#"]
 
     async with aiohttp.ClientSession(
         connector=aiohttp.TCPConnector(verify_ssl=settings.languagetool_verify_ssl)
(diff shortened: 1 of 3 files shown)


app/main.py
```diff
@@ -431,6 +432,13 @@ async def languagetool_rules(user_request_in: RequestIn):
                         for match in result["matches"]:
                             offset = int(match["offset"])
                             end = offset + int(match["length"])
+                            text = user_request_in.text[offset:end]
+                            # ignore text that starts with @ or #
+                            if text[0:1] in ignore or (
```
Szarad, line 437: I can't wrap my head around why this `or` is needed

lsmith77: LT actually also does some spell checking of hash tags .. so if you write `#wrld` they will highlight `wrld` so to filter that out, we also need to check if the sign before is `#` or `@`

Szarad: ah ok, I see now that's for `#`. The test didn't go there, that why I was confused. You didn't want to test `#` as well on one go or another test?

lsmith77: tweaked the test

Szarad (approved):
All good now

PR #323, opened 2022-01-19
improve gendered denom german function to show alternatives even when…

### LINKED ISSUE #313: spacy lemmatiztion issues
* "domiant " (with trailing spaces)
* issues with detecting singular vs. plural

lnazarenko:
… spacy cant define plural or singular form of the word. also to have alt_split in training data for more generic approach for false positives:
- handle empty list for singular/plural form of trigger word to show the whole list of alternatives
- add Alt_split column to training data
- update gendered denom German training data 
- fix test

fixes #313


app/main.py
```diff
@@ -948,52 +949,65 @@ def GenderedDenomAnalysisDE(
...
                 alternative_plur,
+                alternative_all,
                 subcategory,
             ) in gender_words_alternatives:
                 if tokens[i].lemma_ == word:
-                    token_morph_number = tokens[i].morph.get("Number")
-                    if is_number_list_empty(token_morph_number, tokens[i]):
-                        continue
-                    if token_morph_number[0] == "Sing":
+                    if len(tokens[i].morph.get("Number")) == 0:
```
Szarad: Why did you change logic from function is_number_list_empty() to if condition?

app/main.py
```diff
@@ -934,10 +934,11 @@ def GenderedDenomAnalysisDE(
                 word,
                 alternative_sing,
                 alternative_plur,
+                alternative_all,
                 subcategory,
             ) in gender_words_alternatives:
                 if c_doc[i].lemma_ == word:
-                    if c_doc[i].morph.get("Number")[0] == "Sing":
+                    if len(c_doc[i].morph.get("Number")) == 0:
```
Szarad: It is very good you add it here! I missed it when I was writing is_number_list_empty(). But still shouldn't we use here this function as well (most to have logging)?

app/rules.py
```diff
@@ -91,12 +91,13 @@
...
 # df gender
 df_gender = rules["de-DE"]["df_gender_ct"]
-# gender: words + singular alternatives + plural alternatives + subcategory
+# gender: words + singular alternatives + plural alternatives + all alternatives + subcategory
 gender_words_alternatives = list(
     zip(
         df_gender["Lemma"],
-        df_gender["Sg_all_clean"],
-        df_gender["Pl_all_clean"],
+        df_gender["Sg_all_split"],
```
Szarad, line 98: Do we need to have all three columns?  I thought `Alt_split` is sufficient? I just want to double check this with you, you know best.

training_data/de-DE/gendered_noun_words.csv
```diff
@@ -0,0 +1,258 @@
+Lemma,Word_Type,Priority,Category,Primary_subcategory,Sg_all_split,Pl_all_split,Alt_split
```
Szarad, line 1: With that could we remove gendered_noun.csv file?

app/rules.py
Szarad: ah sorry, I didn't check changes in main for `gender_words_alternatives`, this is ok

app/main.py
lnazarenko: yes, we should use something to have logging here

training_data/de-DE/gendered_noun_words.csv
lnazarenko: yes, this file might be remove. I didn't remove it yet because I suppose it used for the false positives PR

app/main.py
lnazarenko: no specific reason, as I wrote you. We can keep is_number_list_empty variable

training_data/de-DE/gendered_noun_words.csv
Szarad: That's true, thank you! I will delete it then when I switch to generic script in generating false positive.

Szarad:
@lnazarenko I push small commit with use of is_number_list_empty() function in if conditions. If these small changes are ok with you, then all good from my side (approving PR).

lnazarenko (commented):
all good from my side

Szarad (approved):
All good!

PR #324, opened 2022-01-21
fix passing in an explicit lang or locale

lsmith77:


app/lang_detection.py
```diff
@@ -61,10 +61,10 @@ def get_locale(self, text, lang, language_preferences, variant_preferences):
 
             return self.get_locale_by_lang(langs, language_preferences)
 
-        if lang in self.supported_langs:
-            if lang in self.supported_locales:
-                return lang
+        if lang in Config._supported_locales:
+            return lang
 
+        if lang in Config.supported_langs:
```
Szarad: you missed underscore ? `Config._supported_langs`

lsmith77: good catch! fixed and added another test

Szarad (approved):
All good now

PR #347, opened 2022-01-25
add a max text length setting

### LINKED ISSUE #338: add limit for maximum text size

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/338

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -262,11 +262,15 @@ async def check_query(
 
     lang = Language(locale)
 
-    list_results = language_rules(user_request_in, lang)
+    text = user_request_in.text
+
+    list_results = language_rules(user_request_in.config, lang, text)
 
     if "orthography" not in user_request_in.config.disabled_categories:
         try:
-            languagetools_results = await languagetool_rules(user_request_in, lang)
+            languagetools_results = await languagetool_rules(
+                user_request_in.config, lang, text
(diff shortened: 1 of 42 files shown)


Szarad (commented):
It is good to add this, but I have 2 concerns:
1. User text could be cut in the middle of the word
2. What with the rest of user's text? User gets some notification?

lsmith77:
> It is good to add this, but I have 2 concerns:
> 
> 1. User text could be cut in the middle of the word
> 2. What with the rest of user's text? User gets some notification?

yeah you are right .. the PR was rushed.

lsmith77:
> 1. User text could be cut in the middle of the word

I now check what the last word is that no longer fits into the limit and cut of there.

> 2. What with the rest of user's text? User gets some notification?

I added something to the response, which the UI could then show to the user

lsmith77:
FTR https://wittyworks.slack.com/archives/C019JFEN4AJ/p1643272554018600?thread_ts=1643211374.015000&cid=C019JFEN4AJ

lsmith77:
FYI I already set the env var on platform.sh to 2000 characters

Szarad (approved):
Thanks for adding extra response and handling text correctly! I played a bit with setting and tested it, looks good for me!

Issue #327, opened 2022-01-21
Line break counting is not correct


arnaumanyosa:
Testing on Twitter, LinkedIn, Witty Form, If I type `\n\nhelo ninja` the highlight position calculation is not correct, see screenshot:

<img width="1080" alt="Screen Shot 2022-01-21 at 14 32 31" src="https://user-images.githubusercontent.com/5278398/150536751-68c218a3-5446-4050-8860-ee414cf877b2.png">

Compare previous result with next one, where I type the same but without the line breaks. The calculated `start` and `end` for the second word in **both cases** is the same, which is wrong and causes the overlap on the UI

<img width="1062" alt="Screen Shot 2022-01-21 at 14 33 49" src="https://user-images.githubusercontent.com/5278398/150536891-d8f555f6-42d7-467b-b905-0e1d4a58a9e6.png">

PR #348, opened 2022-01-26
revert revert in #327, reintroduce replace for linefeeds

### LINKED ISSUE #331: properly handle newlines
API calls with "\n" can lead to not detecting issues, in this example `intern` is not detected

this relates to #295 (initial fix was the replace `\n` with ` `) and #327 (reverted that fix since it was causing issues with start/end)

```
Hi Siwar,\n\nit's the first time that we are looking for a intern\nThomas Jung\nHead of Talent Acquisition EMEA\n\nFossil Group Europe GmbH\nRiehenring 182\n\n4058 Basel\nSWITZERLAND\nwww.fossilgroup.com\nt: +41 61 560 99 69\nm: +41 78 625 82 95\ne: tjung@fossil.com\n\n\n
```

see also https://github.com/explosion/spaCy/issues/4792

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/331

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -455,7 +455,7 @@ async def languagetool_rules(user_request_in: RequestIn, lang: Language):
 
 def language_rules(user_request_in: RequestIn, lang: Language):
     # apply SpaCy pre-built model
-    tokens = model[lang.lang](user_request_in.text.rstrip())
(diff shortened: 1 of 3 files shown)


Szarad (approved):
ah so in the end only `strip()` caused wrong calculation of position. Looks good for me, I tested it, works good.

PR #350, opened 2022-01-27
move fallback form to hubspot

lsmith77:


Szarad (approved):
Great, works good. Maybe when we do some more refactoring url could be put in config file, but I guess it rather won't change.

lsmith77:
yeah .. if we need to change it again .. we can turn it into an env var

PR #351, opened 2022-01-28
287 check alternatives spelling generic

Szarad:
Added generation of ignore file for all training data. Both for German and English.
@lsmith77 let me know if I should add tests for this script. Then I would also need to do refactoring as it is not really possible now (everything is on global scope).


languagetool/English/hunspell/ignore.txt
```diff
@@ -0,0 +1,65 @@
+# words to be ignored by the spellchecker (auto-generated)
```
lsmith77: the ignore.text files now need to be stored in https://github.com/witty-works/languagetool

also relates to the README.md

languagetool/German/ignore.txt
```diff
@@ -0,0 +1,582 @@
+Säger:in
+BeschäftigeR
```
lsmith77: likely a typo in the google sheet.

languagetool/German/hunspell/ignore.txt
```diff
@@ -0,0 +1,66 @@
+# Words to be ignored by the spellchecker (partially generated by SuggestionExtractorTool, also see below)
+# For the opposite (do not accept words), see prohibit.txt.
+# Lines ending with '-*' ignore the previous string only in hyphenated compound words
+einPseudoWortFürLanguageToolTests
+Einpseudowortmitßfürlanguagetooltests-*
+Antoon
```
lsmith77: I kinda wonder if we should sort by alphabet .. maybe this makes reviewing easier?

add_alternatives_to_ignore.py
```diff
@@ -1,156 +1,280 @@
...
                 response = requests.post(
                     api_url,
                     data={
-                        "text": text,
-                        "language": "auto",
-                        "preferredVariants": locale,
+                        "text": word,
+                        "language": locale,
                         "motherTongue": locale,
                         "disabledRuleIds": "SEHR_GEEHRTER_NAME,PROFANITY",
```
lsmith77: BTW I think we should remove `disabledRuleIds` and `disabledCategories` .. we maintain a list of relevant things to ignore over here already https://github.com/witty-works/languagetool/blob/main/server.properties#L22

languagetool/English/ignore.txt
```diff
@@ -0,0 +1,141 @@
...
+mewting
+misuser
+mixologist
+neurotrauma
+neutralise
+newlyweds
+nonelective
+nonverbal
+openminded
+organisation
```
lsmith77: this one is strange, we only have this word in the `en-GB` files (we do have `organization` in the `en-US` files).

so why did this word get added?

I would have thought that setting `motherTongue` as the locale would do the trick here.

lsmith77 (commented):
I added a few comments.

Also informed Tracey about the typos I spotted https://wittyworks.slack.com/archives/C010GGET27M/p1643624370594899

languagetool/English/hunspell/ignore.txt
Szarad: How do you want to do that? I thought generating here and then copying to https://github.com/witty-works/languagetool

languagetool/German/hunspell/ignore.txt
Szarad: words from training data are sorted alphabetically. I think this we shouldn't review because these words were ignored in LanguageTool from the beginning.

languagetool/English/ignore.txt
Szarad: For ignored words for English I added both British English and American English. Should it be just American English, then why is that?

add_alternatives_to_ignore.py
Szarad: ok, great! I remove this

languagetool/English/ignore.txt
lsmith77: I think it makes sense to add both US and GB as long as we use the respective locale to ensure we don't end up adding GB words because they were checked as US. but from reading the code, this is what is already the case, so yeah we might need to leave GB out and simply hope this doesn't result in missing words to add.

languagetool/English/hunspell/ignore.txt
lsmith77: yeah .. that is my thinking too .. but we should then not commit them here imho.

Szarad: Ok, I will open a PR in languagetool repository. Although I would need to leave everything here as well, because he script use "an old ignore.txt" file.

Szarad: But I can add command line argument for ignore file path. Then I could remove languagetool directory from here.

lsmith77: yeah adding a path parameter sounds like the way to go

Szarad:
I had to reintroduce sorting key, because lower-cased words were put after all upper-cased words. Number of words didn't change, only order

PR #352, opened 2022-01-31
for local dev, do a 302 redirect to docs instead of the hubspot form

lsmith77:
I mostly just wanted to prevent the 301 locally since its cached and is a bit annoying if you sometimes also run something else locally on port 8000 .. but it also makes sense to redirect to `/docs` as well since that is the more relevant endpoint locally.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -204,13 +204,20 @@ def openapi(username: str = Depends(get_current_username)):
 
 # public routes
 @app.get("/")
-def get_root():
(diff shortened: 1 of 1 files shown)


lsmith77:
@Szarad you can see that the redirect still works on https://pr-352-cd6kwwq-nfkxhzxe3xgbw.de-2.platformsh.site/

Szarad (approved):
All good, works fine.

PR #363, opened 2022-02-02
minor refactoring to improve code readability

lsmith77:


Szarad (approved):
Looks fine to me. I was a bit stuck when trying to test 500 behavior, but I tested it "manually" with local server, which was down. It was working as it is supposed to.

PR #371, opened 2022-02-07
German rules update

lnazarenko:


lnazarenko:
related to [343](https://github.com/witty-works/nlp_api/issues/343)

PR #377, opened 2022-02-09
fix locale handling in analyze script

lsmith77:


Szarad (approved):
better to change to be sure not to use globals, thanks for the change,

PR #383, opened 2022-02-10
add filters to api responses

### LINKED ISSUE #381: Ability to enable/disable "inspiration alternatives"
As a User I want to be able to remove "inspirations" from the alternatives to just to include the alternatives that will be easy to integrate, so that I don't get overwhelmed or I can make the most use of my time.

<p></p>

Acceptance criteria:

- Users can set the "show_inspiration_alternatives" as a boolean toggle

<p></p>

Design:

<https://www.figma.com/file/oqxFNB3KNmRSJqQnhsrWyq/witty.works-styling?node-id=0%3A1>

### LINKED ISSUE #382: Ability to filter feedback by “importance”
As a User I want to be able to reduce the feedback to just to include the more important highlights, so that I don't get overwhelmed or I can make the most use of my time.

<p></p>

Acceptance criteria:

- Users can set toggle "maximum_importance" as an integer value (1 - highest or 3 - lowest) in the request configuration as "expert mode" (final naming TBD)

<p></p>

Design:

<https://www.figma.com/file/oqxFNB3KNmRSJqQnhsrWyq/witty.works-styling?node-id=0%3A1>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/381
fixes https://github.com/witty-works/nlp_api/issues/382

also prepares for https://github.com/witty-works/browser-extension/issues/287

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -101,6 +101,7 @@ class Config(BaseModel):
     gendered_roles_format: GenderedRolesFormatType = GenderedRolesFormatType.BOTH
     singular_they: str = SingularThey.HE_OR_SHE
     show_inspiration_alternatives: Optional[bool] = False
+    maximum_gravity: Optional[int] = None
 
     @validator("german_gender_ending")
     def valid_german_gender_ending(cls, v: str):
@@ -163,6 +164,7 @@ class ForcedConfig(BaseModel):
     gendered_roles_format: Optional[GenderedRolesFormatType]
     singular_they: Optional[str]
     show_inspiration_alternatives: Optional[bool]
+    minimum_gravity: Optional[int]
 
     @validator("german_gender_ending")
     def valid_german_gender_ending(cls, v: str):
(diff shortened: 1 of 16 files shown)


app/models.py
```diff
@@ -100,6 +100,8 @@ class Config(BaseModel):
     disabled_categories: List = []
     gendered_roles_format: GenderedRolesFormatType = GenderedRolesFormatType.BOTH
     singular_they: str = SingularThey.HE_OR_SHE
+    maximum_gravity: Optional[int] = None
+    show_inspiration_alternatives: Optional[bool] = True
```
lsmith77: maybe we should actually set this to false?

app/models.py
```diff
@@ -318,6 +322,15 @@ def factory(
 
         cleaned_alternatives = []
         for alternative in alternatives:
+            if (
+                not config.show_inspiration_alternatives
+                and ResultOut.isInspirationAlternative(alternative)
+            ):
+                continue
+
+            alternative = alternative.replace("((", "(")
```
lsmith77: as soon as the browser extension supports parsing double parenthesis, these lines need to be wrapped in a client version check

app/models.py
Szarad: Yes, I think it makes sens to set it to `False` .

app/models.py
Szarad: ok, I don't understand why in BE it shows me only 1 inspiration alternatives and when calling API directly 3 (for word `spokesman `)

Szarad: * we can discuss it in a call

Szarad (approved):
Yeah, works good, changes also looks good.

PR #384, opened 2022-02-11
alternatives as object

lsmith77:


app/models.py
```diff
@@ -242,7 +241,14 @@ class RequestInEvent(RequestIn):
     details: Dict[str, str]
 
 
-class ResultOut(BaseModel):
+class ResultAlternative(BaseModel):
+    text: Optional[str]
+    remove: Optional[bool]
+    inspiration: Optional[bool]
+    infos: Optional[str]
```
lsmith77: change this to something else ..

app/categories.py
```diff
@@ -1,391 +1,391 @@
 categories = {
     "abbreviation": {
-        "color": "blue",
         "inclusive": False,
         "category": "style",
-        "gravity": 4,
+        "gravity": 3,
+        "importance": 3,
```
lsmith77, line 6: hide importance in the categories API

app/main.py
```diff
@@ -384,8 +354,50 @@ async def set_rules(user_request_in: RequestIn):
...
 
 
+async def check(
+    version: float,
+    request: Request,
+    response: Response,
+    user_request_in: RequestIn,
+):
+    if version != 1.0 and version != 1.1:
+        response.status_code = status.HTTP_404_NOT_FOUND
```
lsmith77: change to the correct status code

app/models.py
lsmith77: changed to `context`

app/main.py
lsmith77: since we are not using headers for versioning, 406 response code doesn't make sense .. so I just made it a 400 .. right now its impossible to call the check() function with a wrong version anyway. so we could also remove this validation check. I just added it as a pre-caution if we have some accidental changes in the future

app/main.py
```diff
@@ -446,37 +460,48 @@ async def languagetool_rules(config: Config, lang: Language, text: str):
                 result = await r.json()
 
                 list_results = languagetool_matches(
-                    config, lang, "orthography", text, result
+                    version, config, lang, "orthography", text, result
                 )
-            except:
+            except Exception as err:
```
lsmith77: @Szarad I improved the code here to output more information in case of an exception during local development

app/models.py
```diff
@@ -164,7 +165,7 @@ class ForcedConfig(BaseModel):
     gendered_roles_format: Optional[GenderedRolesFormatType]
     singular_they: Optional[str]
     show_inspiration_alternatives: Optional[bool]
-    minimum_gravity: Optional[int]
+    maximum_importance: Optional[int]
```
Szarad, line 168: oh, I haven't seen that before

Szarad (approved):
I tested the gravity feature and inspiration alternatives on both endpoints. From my point of view it looks good.

app/main.py
lsmith77: actually we might just decided to just catch `aiohttp.ClientError` instances?

https://docs.aiohttp.org/en/stable/client_reference.html#client-exceptions

```
diff --git a/app/main.py b/app/main.py
index 5607135..df7ebb1 100644
--- a/app/main.py
+++ b/app/main.py
@@ -462,15 +462,15 @@ async def languagetool_rules(version: float, config: Config, lang: Language, tex
                 list_results = languagetool_matches(
                     version, config, lang, "orthography", text, result
                 )
-            except Exception as err:
+            except aiohttp.ClientError as err:
                 result = "Problem communicating with LanguageTool"
                 if r.status >= 500:
                     try:
                         response = await r.text()
                         result += ": " + response
-                    except:
-                        pass
-                elif settings.platform_environment == "local":
+                    except aiohttp.ClientError as err:
+                        result += ": " + str(err)
+                else:
                     result += ": " + str(err)
 
                 logging.error(result)
```

PR #389, opened 2022-02-15
remove context from response in case of 'store_context' = false

lsmith77:


Szarad (approved):
Works as intended, looks good.

Issue #386, opened 2022-02-15
Add rule level importance, gravity and short explanation, icon


lsmith77:
Each role can optionally customize the (sub-)category properties for:

- importance

- gravity

- short explanation

<p></p>

lsmith77:
Link to feature: https://wittyworks.productboard.com/feature-board/planning/features/11953165

PR #390, opened 2022-02-16
explanation is an object: text, icon, url

lsmith77:


tests/test_general_cases/test_api_english_singular_they/output.json
```diff
@@ -10,7 +10,11 @@
             ],
             "category": "unconscious_bias",
             "end": 14,
-            "explanation": "😟 Pronouns according to binary normative",
+            "explanation": {
+                "icon": "😟",
+                "text": "Pronouns according to binary normative",
+                "url": "https://www.witty.works/en/unconscious_bias#binary_pronouns"
```
Szarad, line 16: this URLs will be soon added? but anyway we are using old version in the browser extension for now?

app/models.py
```diff
@@ -318,11 +325,30 @@ def factory(
             else lang._("rules." + subcategory + "_solution", params)
         )
 
-        explanation = (
-            explanation
-            if explanation != None
-            else lang._("rules." + subcategory + "_explanation", params)
-        )
+        if explanation != None:
```
Szarad, line 328: when this case is happening? That explanation is not None?

Szarad (approved):
I have just two question to clarify things for myself, but works and looks good

tests/test_general_cases/test_api_english_singular_they/output.json
lsmith77: yeah .. the URL format might also change still

app/models.py
lsmith77: this is coming via #386

PR #393, opened 2022-02-16
integrate LT categories as subcategories

### LINKED ISSUE #392: Ability to disable capitalization rules in orthography
category: CASING

rule: UPPERCASE_SENTENCE_START, ..

<p></p>

LT UI

navigation icon changes too

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/392

fundamentally this PR turns LT categories into subcategories of "orthography" so that we can also filter on them.

a side benefit of this is that it also improves the "label" in some cases.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -422,6 +422,20 @@ def languagetool_matches(
                 value = value if value != "" else "-"
                 alternatives.append(value)
 
+        label = match["shortMessage"]
+        if label == "":
+            try:
+                label = match["rule"]["category"]["name"]
+            except KeyError:
+                pass
+
+        try:
+            subcategory = match["rule"]["category"]["id"].lower()
+        except KeyError:
+            subcategory = category
+
+        explanation = match["message"]
+
         list_results.append(
             ResultOut.factory(
                 version,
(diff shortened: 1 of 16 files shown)


app/main.py
```diff
@@ -422,6 +422,20 @@ def languagetool_matches(
                 value = value if value != "" else "-"
                 alternatives.append(value)
 
+        label = match["shortMessage"]
+        if label == "":
+            try:
+                label = match["rule"]["category"]["name"]
```
lsmith77, line 428: sometimes the short message is empty and this ensures we then still have a label to show (rather than just "Orthographie")

app/models.py
```diff
@@ -314,43 +314,46 @@ def factory(
             params["gendered_denominations_ending"] = config.german_gender_ending
 
         label = label if label else lang._("rules." + category + "_label")
-        if category != subcategory:
-            label += ": " + lang._("rules." + subcategory + "_label")
+        if category != "orthography":
+            trans_key = subcategory
```
lsmith77: `trans_key` is not a great variable name .. couldn't think of a better one though :-/

app/main.py
Szarad: yeah, looks better now

app/models.py
Szarad: hmm maybe something like `category_key` ...?

lsmith77: ok .. lets go with that :)

lsmith77: done

PR #396, opened 2022-02-17
refactored german gendering ending code

lsmith77:
to be able to output the list of variations if the analyze script detects issues


Szarad:
@lsmith77 you saw that some tests are failing? Is it only the issue that they were not updated?

Szarad (approved):
very useful:), works good!

PR #397, opened 2022-02-17
fix german gender ending edge cases

lsmith77:


app/models.py
```diff
@@ -533,7 +533,7 @@ def getGenderedRolesFormatInclusive(german_gender_ending, alternative):
 
         if german_gender_ending == "In":
             if alternative.count("~") > 1:
-                ending = ending.capitalize()
+                ending = ending[0:1].capitalize() + ending[1:]
```
lsmith77, line 536: `capitalize()` capitalizes the first char and lower cases all others.

ie. it could lead to `sein:e parter:in` instead of `sein:e Partner:in`

Szarad (approved):
All good now

PR #407, opened 2022-02-23
only capitalize alternatives if the previous non whitespace character is punctuation

### LINKED ISSUE #406: make the auto-capitalization of alternatives smarter
As a User I don't want alternatives capitalized only when it makes sense grammatically so that I don't need to manually fix the capitalization.

<p></p>

Acceptance criteria:

- The current simplistic logic that just looks if the highlighted word was captialized to determined if to automatically captialize the alternatives is replaced with logic that looks if

  - Previous non whitespace character was a punctuation (dot, exclamation mark, questionmark, for german additionally a colon)

  - Previous characters were a double newline

<p></p>

lsmith77:
fixes #406

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -4,6 +4,7 @@ from enum import Enum
 
 import gettext
 import string
+import re
 
 from app.categories import categories
 from app.settings import get_settings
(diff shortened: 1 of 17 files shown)


Szarad (approved):
Works good!

PR #412, opened 2022-02-28
generate categories.csv and subcategories.csv for import into Hubspot

lsmith77:
* [x] generate a categories.csv and subcategories.csv file for Hubspot import
* [x] moved emoji parsing to the importer to reduce the complexity in the models.py


Szarad (commented):
Thanks for clarifying @lsmith77, I will check again the update script.

Szarad:
So the only think I noticed is that for me locally categories.csv `Open discriminating` is missing emoji  this 🚫. I am adding also two minor comments, which you can integrate in this version or another. Besides that looks and works good for me.

update_locales.py
```diff
@@ -124,18 +156,27 @@ def read_csv(in_file):
...
                 )
 
+                if sub_category == categories[sub_category]["category"]:
+                    categories[sub_category]["title"] = {"en": "title", "de": "title"}
+                    categories[sub_category]["text"] = {"en": "text", "de": "text"}
+
                 try:
                     gravity = gravities[str(row[columns["Gravity"]])]
-                except:
+                except Exception:
```
Szarad: Could we be more specific on an exception, like `NameError`?

app/categories.py
```diff
@@ -4,388 +4,482 @@
         "category": "style",
```
Szarad, line 4: hmm I know that we have this file already, but for storing data we should rather use json file not python.

Szarad:
> So the only think I noticed is that for me locally categories.csv `Open discriminating` is missing emoji this no_entry_sign. I am adding also two minor comments, which you can integrate in this version or another. Besides that looks and works good for me.

With new csv file imported from notion I can see missing emoji, but there is nothing in `introduction` column for `openly discriminating`, but afaiu this is expected. From my point of view, PR can be approved.

Szarad (approved):
Works as expected (minor tweaks could be introduced later).

PR #416, opened 2022-03-01
capture context for cases where plural vs. singular cannot be determined

### LINKED ISSUE #415: capture context for cases where plural vs. singular cannot be determined
https://sentry.io/organizations/witty-works/issues/2950555300/?project=5991571&query=is%3Aunresolved&statsPeriod=14d

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/415

unfortunately we don't have a test case .. in fact the aim of this PR is to find a test case :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -319,12 +319,21 @@ async def get_redis(user: str):
 
 
 # Functions
-def is_number_list_empty(number, token):
+def is_number_list_empty(number, token, full_text):
     if not number:
+        start = max(token.idx - 100, 0)
+        end = min(token.idx + 100, len(full_text) - 1)
+        context = full_text[start:end]
+
         logging.error(
-            "List of token morph number: %s for token/word: %s", number, token
+            "List of token morph number: %s for token/word: %s\n%s",
+            number,
+            token.text,
+            context,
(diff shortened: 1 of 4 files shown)


Szarad:
@lsmith77 I committed my proposition for test case (it will fail now, because capture cannot be determined). So should I commented it out for now from test_api.py? or change assert method to "wrong value" and we will update snapshot when it will be handled correctly?

Szarad (changes requested):
@lsmith77 your changes are of course ok, but let's look at proposed test case and how we want to proceed with that.

lsmith77:
I guess adding the wrong value so the test fails with a comment makes sense for now.

Szarad:
I agree, we can always commented it later. I added logging and comment.

lsmith77:
did you remember to push?

Szarad:
yes, I need to figure out how to capture it in github action

Szarad:
*I meant capture in pytest

Szarad:
ok it's fixed now. From my point of view, we can merge and if we decide to disable this test, we can always do it :)

PR #421, opened 2022-03-08
fix max text length handling

lsmith77:
right now the algorithm basically only checks the first paragraph if the text is longer than our limit.

this change here we basically cut the text to `settings.text_max_length` length and then remove the last word to prevent spelling mistakes being detected on the last word if it was cut in half.


app/main.py
```diff
@@ -427,6 +427,12 @@ async def check(
     if settings.read_rules_from_redis:
         await set_rules(user_request_in)
 
+    text = user_request_in.text
```
lsmith77, line 430: moved here since we already want to shorten the text before we do the language detection

app/settings.py
```diff
@@ -19,7 +19,6 @@ class Settings(BaseSettings):
     api_docs_auth_enabled: bool = False
     instrumentation_key: str = ""
     testing: bool = False
-    maximum_text_length: int = 1000
```
lsmith77, line 22: this was never used

Szarad (approved):
Look good, slight change of the output from regular expression to use of `rsplit()`, but it doesn't tear the word which is the most important:).

Issue #374, opened 2022-02-08
show different alternatives for homonyms English depending on part-of-the-speach


lsmith77:
https://github.com/witty-works/nlp_api/issues/254

lsmith77:
Link to feature: https://wittyworks.productboard.com/feature-board/planning/features/11077052

PR #424, opened 2022-03-10
show different alternatives for homonyms English depending on part-of-the-speach

lnazarenko:
related to [374](https://github.com/witty-works/nlp_api/issues/374)
- function type_transform
To transform word_type from rules to SpaCy part-of-the-speech labels
- function homonyms_english
To show different alternatives depending on the part-of-the-speech (word_type) 

Example1: "He took up the challenge.", word challenge is `noun`, alternatives to show:
        "test",
        "burning question",
        "complex situation",
        "task",
        "task requiring thought and skill to solve",
        "complex questions",
        "difficult tasks",
        "burning questions",
        "tasks"
Example2: "My new theory challenges the traditional ideas.", word challenge is `verb`, alternatives to show: 
        "question",
        "cross-examine",
        "re-evaluate",
        "object to",
        "express the need for"

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -773,6 +773,10 @@ def english_rules(version: float, config: Config, lang: Language, tokens, text: 
         sentences_alternatives_en["style"] = style_sentences_alternatives_US
         sentences_alternatives_en["bias"] = bias_sentences_alternatives_US
 
+    list_full += homonyms_english(
+        version, config, lang, text, tokens, rules[lang.locale]["df_homonyms_word"]
+    )
+
     if is_sub_category_enabled(config, "openly_discriminating"):
         list_full += rules_based_words_phrase_matcher_en(
             version,
--- a/app/main.py
+++ b/app/main.py
@@ -738,11 +738,13 @@ def english_rules(version: float, config: Config, lang: Language, tokens, text: 
     gendered_words_alternatives_en = defaultdict(list)
     inclusive_sentences_alternatives_en = []
     sentences_alternatives_en = defaultdict(list)
+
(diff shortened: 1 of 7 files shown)


app/main.py
```diff
@@ -603,6 +603,22 @@ def check_category_importance(config: Config, subcategory: str):
     )
 
 
+# function to convern word_type to SpaCy part-of-the-speech labels
+def type_transform(lemma_type):
+    noun = "NOUN"
```
lsmith77: any specific reason why you define those 4 variables that you only use once?

lnazarenko: In the beginning I thought these variables would be lists of SpaCy labels but then I realised there is no need. I just removed these variables

app/main.py
```diff
@@ -603,6 +603,18 @@ def check_category_importance(config: Config, subcategory: str):
     )
 
 
+# function to convern word_type to SpaCy part-of-the-speech labels
+def type_transform(lemma_type):
+    if lemma_type == "s":
+        return "NOUN"
+    elif lemma_type == "v":
```
lsmith77: small detail the way you had it before was actually better. since you have a return and `if` is actually more appropriate.

note when we move to python 3.10 we could also use a match statement https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching

lnazarenko: Interesting statement. It seems for this simple function type_transform  there is no big benefit to use `match` ... `case `  Interesting articles with cases description https://towardsdatascience.com/the-match-case-in-python-3-10-is-not-that-simple-f65b350bb025

PR #430, opened 2022-03-21
add ability to validate bear auth

### LINKED ISSUE #431: Single Sign On
<https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-register-spa>

<p></p>

<https://github.com/425show/fastapi_microsoft_identity>

<p></p>

[<https://github.com/Intility/fastapi-azure-auth>](https://github.com/425show/fastapi_microsoft_identity)

### LINKED ISSUE #432: apply corporate false positives to all rules
see also <https://github.com/witty-works/nlp_api/pull/279>

<p></p>

probably best to move filtering to just before adding to the results so that we can also collect statistics how often a corporate false positive is triggered

<p></p>

=\> implemented as part of [add ability to validate bear auth #430](https://github.com/witty-works/nlp_api/pull/430)

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/431
fixes https://github.com/witty-works/nlp_api/issues/432

includes https://github.com/425show/fastapi_microsoft_identity/pull/3

also missing tests for now.

getting an access token is a bit painful too.

here is the documentation to get the access token
https://docs.microsoft.com/en-us/azure/active-directory-b2c/access-tokens

but I used postman
https://docs.microsoft.com/en-us/archive/blogs/aaddevsup/using-postman-to-call-the-microsoft-graph-api-using-authorization-code-flow

Auth URL
https://wittyworksdev.b2clogin.com/wittyworksdev.onmicrosoft.com/B2C_1_signupsignin1/oauth2/v2.0/authorize

Access Token URL
https://wittyworksdev.b2clogin.com/wittyworksdev.onmicrosoft.com/B2C_1_signupsignin1/oauth2/v2.0/token

Client ID
7e090308-575a-4e7d-84e4-78c83d083ab6

Client Secret
..

Scope
openid offline_access https://wittyworksdev.onmicrosoft.com/7e090308-575a-4e7d-84e4-78c83d083ab6/access_as_user

State
some random string

https://wittyworksdev.b2clogin.com/wittyworksdev.onmicrosoft.com/B2C_1_signupsignin1/oauth2/v2.0/token

image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -465,22 +465,6 @@ async def check(
 
     if settings.read_rules_from_redis:
         try:
-            organization_object = {
-                "users": ["lukas.smith@witty.works"],
-                "config": {
-                    "forced": {
-                        "german_gender_ending": "In",
-                    },
-                    "suggestion": {},
-                },
-                "false_positive": [
-                    "stark",
-                ],
-            }
-
-            # redis.set("witty.works", json.dumps(organization_object))
-            # redis.set("lukas.smith@witty.works", "witty.works")
-
             claims = validate_scope(settings.aadb2c_expected_scope, request)
             try:
                 await set_rules(user_request_in, claims["emails"][0])
--- a/app/main.py
+++ b/app/main.py
@@ -361,6 +361,7 @@ async def get_redis(user: str, username: str = Depends(get_current_username)):
         if key:
             rules = json.loads(redis.get(key))
             if user in rules["users"]:
+                del rules["users"]
                 return rules
             else:
                 return []
--- a/app/main.py
+++ b/app/main.py
@@ -252,8 +252,19 @@ def german_gender_ending(
     "/auth",
     dependencies=[Depends(HTTPBearer())],
 )
-def auth(request: Request):  # pragma: no cover
-    return validate_scope(settings.aadb2c_expected_scope, request)
+async def auth(request: Request, user_request_in: RequestIn):  # pragma: no cover
+    user = get_user(request)
+    if not user:
+        return user
+
+    claim = validate_scope(settings.aadb2c_expected_scope, request)
+    rules = await set_rules(user_request_in, user)
+
+    return {
+        "claim": claim,
+        "rules": rules,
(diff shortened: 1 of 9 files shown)


app/main.py
```diff
@@ -323,7 +347,6 @@ async def get_redis(user: str):
             user_list = json.loads(redis.get(key))["users"]
```
lsmith77: @Szarad we probably need to do this a bit smarter .. ie. we store additional keys for each email that point to the key of the organization so that we don't need to iterate over all the keys.

tests/test_api.py
```diff
@@ -305,36 +302,18 @@ def set_redis():
...
     # Set a value
     redis.set("test", json.dumps(organization_object))
+    redis.set("test@gmail.com", "test")
 
 
 @pytest.mark.parametrize(
     "fp_case_dir",
     get_dirs("tests/test_false_positive"),
 )
 def test_false_positive(fp_case_dir, snapshot, set_redis):
```
lsmith77, line 312: we need to have some way to inject a bearer token that points to `test@gmail.com`

Szarad:
I successfully authenticate, added rules for organization, displayed rules, but it doesn't forced the organization config to me (but trying to test it now).

Szarad:
So I have this corporate rules (German ending is forced on `:in`)
corporate_rules
But with this request (user set endings to be In):
`{
  "type": "check",
  "text": "Lernpate",
  "lang": "de-DE",
  "config": {
    "german_gender_ending": "In"
  }
}`
results are : "LernpatIn", not `Lernpat:in`
We need to check where the problem lies :/, when I was implementing it the first time it was working, but since then there was a lot of changes.

lsmith77:
one note .. I was testing the case when the forced rule is used to override an API default. I didn’t test the case when the forced rules override an explicitly passed in value. then again I don’t think that should have an influence on this logic.

Szarad (approved):
Code looks good, more test with dashboard needed.

PR #435, opened 2022-03-24
re-add fasttext model caching

lsmith77:


Szarad (approved):
Make sens to reintroduce it from your conversation with Nils.

PR #438, opened 2022-03-29
add support for storing term replacements

lsmith77:


app/models.py
```diff
@@ -220,12 +220,26 @@ def split_string_values(cls, v):
...
+    explanation: Optional[Explanation]
+    gravity: Optional[int]
+
+
 class ConfRequest(BaseModel):
     organization: str
     users: list
     forced: ForcedConfig
     suggestion: Config
-    false_positive: Optional[list] = []
```
lsmith77, line 228: renamed for consistency

Szarad (approved):
I tested it, works goo, code also look good to me. One minor comment is that I don't see the term_replacements structure in docs. It would be nice to have: ` "term_replacements": [{"term":"string", "alternatives": ["string"], explanation:{"text":"string", "icon":"string", "url":"string"}, "gravity":"int"}]`

lsmith77:
yeah not sure how to achieve that .. however if you click on schema you get more details

image

PR #300, opened 2022-01-17
scrub sentry data with presidio

lsmith77:
fixes #297

https://presidio-demo.azurewebsites.net/
https://microsoft.github.io/presidio/analyzer/

This implementation works fine but note the list of supported entities is limited:
https://microsoft.github.io/presidio/supported_entities/

see also https://www.notion.so/witty-works/Privacy-vs-Data-Collection-1dbf896818cf4067a8b932a4e3163829

example see https://sentry.io/organizations/witty-works/issues/2975885227/?environment=local&project=5991571&query=is%3Aunresolved

consider adding additional regexp:
* [ ] remove all numbers "words" (ie. `\s\d+\s`)
* [ ] remove all words with only numbers and uppercase letters (ie. `\s[\dA-Z]+\s`)
* [x] https://github.com/microsoft/presidio/issues/822

PR #449, opened 2022-04-22
filter out emails, urls and numbers in context and sentry

### LINKED ISSUE #297: Remove PII from Context and Sentry logs
We need to filter out the text parameters from Sentry logs (both within the API also when communicating with LanguageTool)

https://docs.sentry.io/platforms/python/guides/logging/data-management/sensitive-data/

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/297
supersedes https://github.com/witty-works/nlp_api/pull/300

code is based on https://github.com/lmeulen/PrivacyFilter
we will have to see what the impact on performance will be but I think it shouldn't be too much.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -8,6 +8,7 @@ import re
 
 from app.categories import categories
 from app.settings import get_settings
+from app.privacy_filter import get_privacy_filter
 
 
 class Language(object):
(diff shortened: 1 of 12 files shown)


lnazarenko (approved):
Very important feature to remove all sensitive information from user's text. You should write this to data user's agreement or something similar because users worry about this. 
I don't think it should influence on performance because Python regular expression library is time of the fastest for text handling

lsmith77:
yeah .. once we release this I will update https://www.witty.works/trust-and-security

PR #450, opened 2022-04-23
English-US rules

lnazarenko:


lnazarenko (commented):
the problem was with missing value for the subcategory in google sheet.

lsmith77 (approved):

PR #468, opened 2022-05-13
Update the generate_false_positive.py script and false positives list.

Szarad:
Update generate_false_positive.py script to be up to date with current NLP API code. Run the script to update list of false positives.


bin/generate_false_positive.py
```diff
@@ -73,32 +73,6 @@ def check_if_server_is_running(api_url):
 for file in os.listdir(base_directory):
     training_data_paths.append(base_directory + file)
 
-
```
lsmith77, line 76: why did you remove these?

Szarad: These lines weren't used, so to make the code more readable.

bin/generate_false_positive.py
```diff
@@ -130,7 +104,7 @@ def check_if_server_is_running(api_url):
     if "~" in word:
         for german_gender_ending in endings:
             alternative = ResultOut.getGenderedRolesFormatInclusive(
-                word, german_gender_ending
+                german_gender_ending, word
```
lsmith77, line 107: ooops .. good catch

PR #469, opened 2022-05-13
Update docker build action

Szarad:
Add step to update requirements,txt file for the Docker image.


lsmith77 (changes requested):
I guess we can then delete https://github.com/witty-works/nlp_api/blob/dev/requirements.txt ?

Szarad:
> I guess we can then delete https://github.com/witty-works/nlp_api/blob/dev/requirements.txt ?

Yes, that's right. I removed this file.

lsmith77:
ok great .. can you also review the README.md ?

Szarad:
> ok great .. can you also review the README.md ?

Yes, I updated it accordingly.

PR #471, opened 2022-05-18
added special handling to keep Express.js etc together

lsmith77:
relates to https://www.notion.so/witty-works/Amend-issues-in-Ignore-texts-May-2022-1654b4d16a8143ada89f26f72cb9b66e


Szarad (approved):
Looks and works good. Just one small comment, why is the replacement string that complex? Wouldn't be enough to replace `.js` with `~js`?

lsmith77:
no specific reason .. just to be extra sure

PR #477, opened 2022-06-06
handle -en adj/adv

### LINKED ISSUE #476: handling of "en" words
see <https://www.notion.so/witty-works/Are-there-exceptions-in-case-of-en-ending-to-be-taking-from-terms-to-alternatives-c68f2475785f47dba5984cd55d305c75>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/476

just poking around here .. likely the transformation would have to be added to other functions too


app/main.py
```diff
@@ -1089,6 +1089,35 @@ def ing_ify_alternatives(token, alternatives):
...
+"""Function to change adjectives to -en form in alternatives"""
+
+
+def en_ify_alternative(lang, alternative):
+    tokens = model[lang.lang](alternative.rstrip().replace("\n", " "))
+
+    alternative = ""
+    for token in tokens:
+        alternative += token.text
+        if token.text[-1] != "m" and (
```
lsmith77: this prevents `mit gutem Ergebnis` to be converted .. not sure if this isn't more an issue with this alternative just not fitting like the others

lnazarenko (commented):
If the "en" adding to alternatives should be for the verbs as well? 
Here:  if token.pos_ == "ADJ" or token.pos_ == "ADV" or token.pos_ == "VERB":

lsmith77:
yeah .. added ..

what then remains is deciding in which of the functions we need to apply this on alternatives?

lnazarenko (approved):
It seems now is good:
en_ify_alternative function adds "en" to single-click alternatives if there are adjectives, adverbs and verbs. 
en_ify_alternatives function changes alternatives by adding "en" if user's words ends with "en" and these words are adjectives, verbs and adverbs

lnazarenko:
This type of alternative transformation needs to be added to ub_words_phrase_matcher_de and it will cover all adjectives, verbs and adverbs from google sheet

lsmith77:
@lnazarenko I had to do some more adjustments but now it seems to be correct. can you check one more time?

lsmith77:
note I don't see anything we can do to make it pass the quality gate. in fact I think its a bug in sonar, since this PR doesn't introduce the duplicate blocks sonar complains about. it just adds to those blocks.

lnazarenko (commented):
Why for the word "Kunden" there is empty list of alternatives? I mean this test: Wir suchen Ninja Programmierer für unseere Kunden
It should be some single click alternatives for this word

lsmith77:
FTR I spend a bit of time to reduce code duplication in general, which is always a good idea anyway :)

the list is empty for `Ninja`

https://github.com/witty-works/nlp_api/blob/f62fc305a5b7a1fc343181c112fe848ef4b61e1c/tests/test_1_0/test_api/output.json

lnazarenko:
Ah, yes, right. It was three inspiration alternatives for Ninja before and now the API doesn't show them

lnazarenko (approved):

PR #486, opened 2022-06-14
hotfix to support alternate german gender ending syntax

lsmith77:
the browser extension is sending `colon_in` instead of `:in`
so I am now allowing this representation but I map it to the proper syntax.


arnaumanyosa (approved):
LGTM 🚀

PR #502, opened 2022-06-21
validate azure ad b2c client id

lsmith77:


hellasol (approved):
Looks good! Just one question: When I recieve a 403, I try to get a new token using the refresh token. Will this process also fail (from the backend) if the refresh token does not match the current user?

lsmith77:
> Looks good! Just one question: When I recieve a 403, I try to get a new token using the refresh token. Will this process also fail (from the backend) if the refresh token does not match the current user?

correct

PR #506, opened 2022-06-24
update german lemmatizer

### LINKED ISSUE #505: starke Schiene
We don't detect "starke" even though "stark" is in our rules "Deutschland braucht eine starke Schiene."

lnazarenko:
fixes #505

### DIFF
--- a/app/model.py
+++ b/app/model.py
@@ -187,6 +187,7 @@ dict_lemma_lookup_de = {
     "schlicht": "schlicht",
     "sicher": "sicher",
     "stärker": "stärker",
+    "starke":"stark",
     "treibend": "treibend",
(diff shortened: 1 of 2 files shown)


lnazarenko:
the word `starke` was not highlighted due to the known problem with SpaCy lemmatiser.
Updated custom lemmatizer with this word
We saw the same problem with female German adjectives last year with the word `internationale`

lnazarenko:
Question regarding tests. Now the test with corporate false positives failed because it started to highlight starke in this sentence "Ich bin eine starke Führungskraft!"  And it is not clear for me should or shouldn’t starke be highlighted in such case with corporate false positives

lsmith77 (approved):

PR #507, opened 2022-06-27
add s in english, f.e. 'decides' => 'accepts'

lsmith77:
also ensure that we only modify token's that match the original token type (ie. verb or adjective) in english, and consider "PROPN" to be the same type if there is only one token in the alternative

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -513,13 +513,15 @@ def get_user(request: Request):
         try:
             if claims["aud"] != settings.aadb2c_client_id:
                 raise HTTPException(
-                    status_code=status.HTTP_403_FORBIDDEN, detail="access token does not match client id"
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="access token does not match client id",
                 )
 
             return claims["emails"][0]
         except KeyError:
             raise HTTPException(
-                status_code=status.HTTP_403_FORBIDDEN, detail="access token does not map to email"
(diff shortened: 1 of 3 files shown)


lsmith77:
to be discussed in https://www.notion.so/witty-works/s-ending-on-english-5e2a4c3800494cdb98f15b3547842691

lnazarenko (approved):

app/main.py
```diff
@@ -1203,19 +1214,35 @@ def alternative_declension(text, ending, lang, alternative):
...
+        if previous == False:
+            alternative_token_type = get_token_type(
+                token, token_type, len(tokens) == 1
+            )
+
+            if lang.lang == "en":
+                if  "verb" == alternative_token_type:
+                    previous = True
+
+                    if text[-1] == "h":
```
lsmith77: are there other cases we need to consider here?

app/main.py
```diff
@@ -1203,19 +1214,35 @@ def alternative_declension(text, ending, lang, alternative):
...
     return new_alternative
 
 
 def alternatives_declension(token, lang, alternatives):
-    if adjective_or_verb(token):
+    endings = False
+    token_type = get_token_type(token)
+
+    if lang.lang == "en" and token_type == "verb":
+        endings = ["s"]
```
lsmith77, line 1244: are there other relevant endings?

PR #509, opened 2022-06-27
check word type when matching a lemma

### LINKED ISSUE #508: False positive "überlegen"
<p></p>

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/508


lsmith77:
I would expect that we actually match the word type in all our rule functions.

however oddly enough for "Wir sind überlegen" spacy incorrectly detects "überlegen" as a verb and not as an adjective.

lsmith77:
I would expect that we actually match the word type in all our rule functions.

however oddly enough for "Wir sind überlegen" spacy incorrectly detects "überlegen" as a verb and not as an adjective. same for "herausgefordert" in "Wir lieben es, herausgefordert zu werden."

lsmith77:
@lnazarenko do you have an idea here?

lsmith77:
closing in favor of https://github.com/witty-works/nlp_api/pull/518

PR #515, opened 2022-07-05
re-start declension if a conjunction is encountered

lsmith77:
fixes https://github.com/witty-works/nlp_api/runs/7192949211?check_suite_focus=true


lsmith77:
BTW in theory spacy can detect conjunctions but when I tried it, it didn't work. probably because we would need a larger model:
https://stackoverflow.com/posts/63309539/revisions

lnazarenko (approved):

PR #518, opened 2022-07-06
german-word-type-uberlegen

### LINKED ISSUE #508: False positive "überlegen"

lnazarenko:
fixes https://github.com/witty-works/nlp_api/issues/508

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -794,18 +794,17 @@ async def language_rules(
 
 
 # Function for all German rules
-
-
-def is_not_noun(pos):
-    return pos != "NOUN" and pos != "PROPN" and pos != "PRON"
-
-
-def get_non_noun_lower_cased(token):
+def get_lemma_lower_cased(token):
     token_word = token.lemma_
-    if is_not_noun(token.pos_):
-        token_word = token_word.lower()
 
-    return token_word
+    return token_word.lower()
+
+
+def get_lemma_non_noun_lower_cased(token):
+    if token.pos_ != "NOUN" and token.pos_ != "PROPN" and token.pos_ != "PRON":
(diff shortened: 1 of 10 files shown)


lnazarenko:
fix this [508](https://github.com/witty-works/nlp_api/issues/508) But case like "wir sind überlegen" needs transformer-based model from SpaCy to make the part-of-the-speech tags more precisely

lnazarenko:
I added a list of possible part-of-the-speech-tags to recognise words as adjectives or adv. It helps to fix an issue with "herausgefordert" in "Wir lieben es, herausgefordert zu werden." As it was recognise as perfect partciple which is correct in such case

lnazarenko:
It works with small model but with larger model it works more precisely, as well as larger model helps to recognise plural or singular forms much better

tests/test_demo_wordings_german/test_internal_email/output.json
```diff
@@ -231,21 +231,6 @@
             "subcategory": "agentic",
             "text": "herausgefordert"
         },
-        {
```
lsmith77, line 234: `lieb,a,inclusive,communal` but here `lieben` is a verb

app/main.py
```diff
@@ -1639,7 +1653,7 @@ def style_word_analysis_de(
     patterns = [model[lang.lang].make_doc(text) for text in terms]
     matcher.add("TerminologyList", patterns)
 
-    for i in range(len(tokens))[1:-1]:
```
lsmith77, line 1642: why did we skip the first and last token?

lsmith77 (approved):

PR #527, opened 2022-07-12
ignore term replacement alternatives

### LINKED ISSUE #526: make all term replacement alternatives implicit ignore words
otherwise people might select an alternative they configured and it would then be highlighted as spelling mistake or marked as non inclusive.

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/526

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -477,8 +477,9 @@ def merge_rules(user_request_in: RequestIn, organization_rules: list):
         data = organization_rules["config"][config]
         if data is not None and data["status"] == "force":
             if config in categories:
-                if data["value"] and config in disabled_categories:
-                    disabled_categories.remove(config)
+                if data["value"]:
+                    if config in disabled_categories:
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -477,8 +477,9 @@ def merge_rules(user_request_in: RequestIn, organization_rules: list):
         data = organization_rules["config"][config]
         if data is not None and data["status"] == "force":
             if config in categories:
-                if data["value"] and config in disabled_categories:
-                    disabled_categories.remove(config)
+                if data["value"]:
```
lsmith77, line 480: this is an independent bug fix.

when merging organization rules with the request config, we always disabled all categories that didn't exist in the request config. this luckily doesn't affect the browser, since the browser always included explicit settings for each of the disabled categories. so this bug only affected other uses of the API, f.e. in our tests.

lnazarenko (approved):

PR #532, opened 2022-07-18
only filter the number of a number is contained within a string

lsmith77:
I stumbled over this as I was creating a test for a hashtag containing a number and it replaced the entire hashtag and not just the number, which was the intention of the privacy filter.

### DIFF
--- a/app/privacy_filter.py
+++ b/app/privacy_filter.py
@@ -56,7 +56,7 @@ class PrivacyFilter:
         )
 
     def remove_numbers(self, text):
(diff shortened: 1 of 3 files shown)


app/privacy_filter.py
```diff
@@ -74,7 +74,7 @@ def filter_regular_expressions(self, text):
         text = self.remove_email(text)
         text = self.remove_url(text)
         text = self.remove_numbers(text)
-        return text
+        return text.strip()
```
lsmith77, line 77: this is the appropriate place for the `strip()` call

lnazarenko (approved):

PR #537, opened 2022-07-19
add article to german gender denom alternatives

### LINKED ISSUE #535: Handle "der Kaufmann" as one term
Rather than analyzing "der" (ein etc) and "Kaufmann" separately, we should handle them as one term, f.e. suggesting "die:der Kauffrau:Kaufmann"

<p></p>

He is doing great. (1st click) =\> They is doing great. (2nd click) =\> They are doing great

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/535


lsmith77:
@lnazarenko unfortunately it is more complicated since we sometimes re-frame (ie. Mitarbeit) or pluralize (ie. Personal). I tried to detect plural alternatives and then change it to `die` but that didn't work.

app/gender.py
```diff
@@ -0,0 +1,182 @@
...
+
+    if word.endswith("t"):
+        return {
+            "word": word,
+            "definite_article": "die",
+            "indefinite_article": "eine",
+            "warning": "Unsure: Most nouns ending in -t originating from verbs are feminine.",
+        }
+
+    if word.endswith("in"):
```
lsmith77: https://github.com/Pimverleg/German-gender-finder/issues/2

app/gender.py
```diff
@@ -0,0 +1,182 @@
+import sys
```
lsmith77: https://github.com/Pimverleg/German-gender-finder/issues/1

lnazarenko (approved):

PR #554, opened 2022-08-22
handle false positives patterns

### LINKED ISSUE #523: handle the false positive column
exists in english and german

<p></p>

lnazarenko:
fixes https://github.com/witty-works/nlp_api/issues/523


lnazarenko:
Finally! This part is handle false positives from the false positives column in witty rules (need to extend to handle abbreviations). words in multi terms. This part might be extended. 
The overall: there are two function to find false positives in the user text, one based on exact match, another one on patterns. If token is a trigger word then check that the position of this token is outside of the range of the false positive multi terms. Depending on this check is True or False, this trigger word goes to output or not.

app/main.py
```diff
@@ -2198,8 +2198,55 @@ def rules_based_words_phrase_matcher(
...
+
+# matcher to false positives
+def is_false_positive_match(list_false_positive, tokens, token):
+    if list_false_positive.__len__() > 0: 
+        for match_id, start, end in list_false_positive:
+            span_false = tokens[start:end]
+            if token.idx in range(span_false.start_char, span_false.end_char):
+                return True
+        return False  
+    else:
```
lsmith77: remove `else`

app/main.py
```diff
@@ -2213,28 +2260,33 @@ def rules_based_words_phrase_matcher_en(
 ):
     list_tokens = []
 
+    #create false positives list 
+    phrase_matches_false = get_matches(tokens, list_false_column)
```
lsmith77: move this into a function `get_false_positive_matcher` and call it only once and then ad `matches_false` as a parameter in all the functions that have the `if is_false_positive_match(matches_false, tokens, token) == False:` condition

PR #557, opened 2022-08-23
handle notification count

### LINKED ISSUE #556: show notification if there is a team invitation
As a User I want to see a notification in the browser if there is an open team invitiation to accept on the dashboard.

<p></p>

Acceptance criteria:

- In the nav menu, show a notification if there is an open navigation

- Change the nav bar icon to show the user they should open the nav menu

API is missing for this. I guess I will integrate this into the NLP API somehow but I have not done so yet .. but I guess I would set “config_changed” = true and then add something to the /auth endpoint .. does that make sense for you?

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/556

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -302,6 +302,7 @@ class ConfRequest(BaseModel):
     term_replacements: Dict[str, TermReplacement] = {}
     domains: Optional[DomainConfig]
     config_hash: Optional[str]
+    notifications: Optional[int]
 
 
 class ConfRequest1_1(BaseModel):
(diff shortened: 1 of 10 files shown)


app/models.py
```diff
@@ -331,7 +332,7 @@ class RequestIn(BaseModel):
     client: Optional[str] = None
     config: Optional[Config] = Config()
     config_hash: Optional[str]
-    config_organization_hash: Optional[str]
+    organization_config_hash: Optional[str]
```
lsmith77, line 335: @hellasol notice that I changed the API here slightly so that the naming is consistent `organization_config_hash`

app/models.py
```diff
@@ -793,3 +794,4 @@ class ResultsOut(BaseModel):
     language: str
     limit_reached: bool
     config_changed: Optional[bool]
+    notifications: Optional[int]
```
lsmith77, line 797: I am returning the number of `notifications` as part of the check response now

app/models.py
hellasol: noted!

hellasol (approved):

PR #567, opened 2022-09-02
English offensive language and rules update

### LINKED ISSUE #536: offensive language
add to rule-editor and API

lnazarenko:
fixes https://github.com/witty-works/nlp_api/issues/536


lsmith77:
@lnazarenko did this also require changes to the rule editor?

lnazarenko:
offensive language is subcategory - it doesn't require to change the rules editor.
The only change is in alternatives: there is one more column "Identity_first" for 18 words.

lsmith77:
so you did this export without the rule editor or you have modified the rule editor accordingly already?

lnazarenko:
I did export without rules editor. You can create a ticket for the rules editor separately if you want.

lsmith77:
ah ok .. it is very important to me that we only use the rule editor to export rules. there are minor differences like `['']` vs `[]` in the output and it will get confusing.

more importantly the rule editor ensures that we can keep track of exactly what code we use for exports.

so imho we always need to start with adapting the rule editor. I will then create explicit tickets for this now.

lsmith77:
closing in favor of https://github.com/witty-works/nlp_api/pull/573

Issue #568, opened 2022-09-05
ensure case-insensitive match


lsmith77:
`first class citizen` vs. `First class citizen`

![IMG_0667](https://user-images.githubusercontent.com/300279/188401095-2f82d0fc-b4b1-4a83-97da-dacdd55fae8e.jpeg)

lsmith77:
Link to feature: https://wittyworks.productboard.com/feature-board/planning/features/15194115

lsmith77:
I had a quick look at this issue to see if all we needed to do is replace `if span.text == sentence:` with `if span.text.lower() == sentence.lower():`.

But the issue is that the phrase matcher already doesn't match `First class`.

```
def get_matches(tokens, phrases):
    # Phrase matcher part to handle False positives with two words and special symbols
    matcher = PhraseMatcher(model[lang.lang].vocab)

    # Only run model.make_doc to speed things up
    patterns = [model[lang.lang].make_doc(text) for text in phrases]
    matcher.add("TerminologyList", patterns)
    return matcher(tokens)
```

PR #574, opened 2022-09-08
fix upper and lower case match for multiterms

lnazarenko:
https://github.com/witty-works/nlp_api/issues/568


lsmith77:
@lnazarenko there are 3 more cases where we have `if span.text == sentence:` .. should those also be updated?

lsmith77 (approved):

PR #578, opened 2022-09-10
removed support for the old 1.1 format and endpoints for rules stored in redis

lsmith77:
this also cleans up the code in other ways:

* merged several of the rule functions to reduce code duplication
* removed redundant code communicating with redis
* aligned function names
  * `fetch` is used for internal functions (so that `get`, `delete`, `post` etc can be used for routes)
  * `_from_redis` reads from redis
  * `_from_request` uses the request data to then read from relevant data sources (ie. redis)

also dropped the `/form` endpoint since it is for quite some time no longer used and redirect to `/editor` and not `/form`

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -756,6 +756,32 @@ async def fetch_organization_rules_for_request(
     return rules
 
 
+def fetch_email_from_claims(claims):
+    try:  # pragma: no cover
+        if claims["aud"] != settings.aadb2c_client_id:
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail="access token does not match client id",
+            )
+
+        if "email" in claims:
+            return claims["email"]
+
+        if "emails" in claims and len(claims["emails"]) > 0:
+            return claims["emails"][0]
+
+    except KeyError:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="access token does not map to email",
+        )
+
+    raise HTTPException(
+        status_code=status.HTTP_403_FORBIDDEN,
+        detail="no email found in claim",
+    )
+
+
 def fetch_user(request: Request):
     if "authorization" in request.headers and request.headers[
         "authorization"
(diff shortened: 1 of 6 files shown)


hellasol (approved):

app/main.py
```diff
@@ -1593,7 +1504,7 @@ def english_rules(version: float, config: Config, lang: Language, tokens, text:
         )
 
     if is_sub_category_enabled(config, "inclusive"):
-        list_full += rules_based_words_phrase_matcher_no_alt_en(
+        list_full += rules_based_words_phrase_matcher_en(
```
lnazarenko: rules_based_words_phrase_matcher_no_alt_en doesn't provide alternatives, currently it is used only for inclusive category
rules_based_words_phrase_matcher_en this is common function for English rules and it provides alternatives

app/main.py
```diff
@@ -1857,7 +1772,8 @@ def plural_or_singular_en(
 ):
     if token_morph_number[0] == "Sing":
         return alternative_sing, subcategory
-    elif token_morph_number[0] == "Plur":
+
```
lnazarenko, line 1833: why did you change elif to if? Usually in Python if.. elif... else... statements

lsmith77: because the previous `if` does a return

app/main.py
lsmith77: yeah .. I have adapted `rules_based_words_phrase_matcher_en` to be able to handle the case when no alternatives are provided. I can show you how this works on Thursday

app/main.py
```diff
@@ -2407,11 +2331,10 @@ def rules_based_words_phrase_matcher_de(
 
 # matcher to false positives
 def is_false_positive_match(list_false_positive, tokens, token):
-    if list_false_positive.__len__() > 0:
```
lnazarenko, line 2404: why did you remove the check of the length of false positives list?

lnazarenko (approved):

app/main.py
lsmith77: because if the list is empty, then the for loop will just not iterate over anything

PR #583, opened 2022-09-20
update sentry integration, also update all dependencies

### LINKED ISSUE #569: check why sentry is not receiving 500 error stack traces

lsmith77:
Sentry just released official support for FastAPI, so we can remove a lot of custom code:
https://blog.sentry.io/2022/09/01/fastapi-and-starlette-sentry-integrations-have-arrived/

fixes https://github.com/witty-works/nlp_api/issues/569

### DIFF
--- a/app/sentry.py
+++ b/app/sentry.py
@@ -1,5 +1,8 @@
 import sentry_sdk
 from sentry_sdk.integrations.aiohttp import AioHttpIntegration
+from sentry_sdk.integrations.starlette import StarletteIntegration
+from sentry_sdk.integrations.fastapi import FastApiIntegration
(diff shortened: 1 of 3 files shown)


app/sentry.py
```diff
@@ -13,7 +16,7 @@ def sentry_clean_sensitive_frame(
     return frame
 
 
-def sentry_clean_event_data(event):  # pragma: no cover
+def sentry_clean_event_data(event, hint):  # pragma: no cover
```
lsmith77, line 19: it seems this was actually the root issue

hellasol (approved):

PR #599, opened 2022-09-26
use mailjet to send statistics email

### LINKED ISSUE #598: use Mailjet for emails
<https://github.com/mailjet/mailjet-apiv3-python>

<https://github.com/witty-works/nlp_api/blob/dev/send_statistics.php>

<p></p>

[<https://gist.github.com/bsartek/3808dc6a7f214cf2b4bcae7e36d38fe8>](https://github.com/witty-works/nlp_api/blob/dev/send_statistics.php)

[[](https://gist.github.com/bsartek/3808dc6a7f214cf2b4bcae7e36d38fe8)](https://github.com/witty-works/nlp_api/blob/dev/send_statistics.php)

drop sendgrid from SPF

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/598

the goal is essentially that we are only sending emails via google, mailjet and hubspot to make it easier to manage our spam ratings.


.platform.app.yaml
```diff
@@ -53,8 +53,8 @@ relationships:
 
 crons:
   generate-statistics:
-    spec: "55 0 * * *"
+    spec: "0 1 * * 0"
```
lsmith77, line 56: once a week is enough

statistics.sh
```diff
@@ -62,13 +72,51 @@ do
...
+          eval $cmd;
+        fi
     fi
 
 done
+
+mj_payload(){
+    currentDate=`date +"%Y-%m-%d"`
+    messageJson=`echo "$message" | jq -Rsa .`
+    messageJson=${messageJson//\\\\/\\}
```
lsmith77, line 97: not quite sure why this is needed .. but otherwise some of the `\n` end up double escaped.

hellasol (approved):

PR #607, opened 2022-10-03
ability to use the v2 API without logging in

### LINKED ISSUE #606: show highlights when not logged in
As a User I want to see why I should log in and what I am missing if I am not logged in\.

<p></p>

Acceptance criteria:

- When not logged in, still highlight words in grey

- In the popover show the same content as in the popup until the user has logged in

<p></p>

Design:

<p></p>

Feel free to adapt the design Solveig if needed\. I created two different versions\. You can find the design here\.

<p></p>

<p></p>

Implementation Thoughts:

- On the API we need to support non\-logged in requests against the \/v2\/check endpoint

- In this case return `premium` as the \(sub\-\)category without any alternatives

- Frontend then essentially only needs to use the new highlight color and custom content in the popover \(similar to what we have in the popup when the user is not logged in\)

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/606

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -552,6 +552,42 @@ async def post_check_v2_0(
     )
 
 
+@app.post(
+    "/v2.1/check",
+    response_model=Union[ResultsOut, Result],
+    response_model_exclude_none=True,
+    dependencies=[Depends(HTTPBearer(auto_error=False))],
+)
+async def post_check_v2_1(
+    request: Request,
+    response: Response,
+    user_request_in: RequestIn,
+):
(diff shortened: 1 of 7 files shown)


hellasol (approved):
when you merge this to dev, the extension won't work because of validation errors. Do you want to merge and then I push the fix?

lsmith77:
well then maybe the better approach is to do all of this only for a v2.1 check endpoint

hellasol:
> 

yes, good idea!

lsmith77:
ok done

hellasol (approved):

PR #625, opened 2022-10-17
fix conjunction detection

lsmith77:


app/main.py
```diff
@@ -1861,17 +1861,6 @@ def add_declension_german(text, ending):
     return text + ending
 
 
-def is_conjunction(full_text, start=0):
-    if full_text in ["und", "oder", "and", "or"]:
-        return True
-
-    preceeding_text = full_text[max(0, start - 5) : start]
```
lsmith77, line 1868: this code never really belonged here as it was just relevant for the edge case of `aber` used as a filler word and not as a conjunction

lnazarenko (approved):

PR #631, opened 2022-10-18
return only the top 5 alternatives

### LINKED ISSUE #630: limit the number of alternatives returned
since we only show 5 anyway, why return more?

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/630


hellasol (approved):
you have some conflicts, but except from that it looks good : 👍

PR #636, opened 2022-10-25
match article in more cases, remove article when pluralizing

### LINKED ISSUE #633: detect proper grammatical case
"In unserer Mannschaft sind wir sehr ambitioniert"

<p></p>

The case has to be changed for each of the alternative nouns.

“In unserem Team…”

“In unserer Equipe…”

“In unserer Gruppe…”

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/633


lsmith77:
further idea is to leverage the article and noun information to determine what case it is (ativ, genetiv, akkusativ, nominativ).

for nouns we get information via german-nouns https://github.com/gambolputty/german-nouns#lookup-words

for articles we would need to add the relevant information to the articles.csv:
https://wittyworks.slack.com/archives/C010GGET27M/p1666692078976009

tests/test_grammatically_correct_alternatives/test_german_boss/output.json
```diff
@@ -0,0 +1,38 @@
+{
+    "language": "de",
+    "limit_reached": false,
+    "results": [
+        {
+            "alternatives": [
+                {
+                    "text": "Unser Leitungsperson"
```
lsmith77, line 8: this here is incorrect because our library does not detect the gender of `Leitungsperson` properly as a result we just keep the article as in the original text

lnazarenko (approved):

PR #650, opened 2022-11-11
do not return team_analytics, instead hide the organization_id

lsmith77:


app/main.py
```diff
@@ -578,18 +583,13 @@ async def post_check_v2_1(
...
-        team_analytics = rules["team_analytics"]
-
     return ResultsOut(
         results=results,
         language=language,
         limit_reached=limit_reached,
         config_changed=fetch_config_change(rules, user_request_in),
         notifications=notifications,
         has_consented_to_mailing=has_consented_to_mailing,
-        team_analytics=team_analytics,
```
hellasol, line 592: is this supposed to be removed?

lsmith77: yes .. since we are removing the `organization_id` in case the user has opted out of `team_anaytics`, we no longer need to return the `team_analytics` to the browser extension.

hellasol: aha.. ofc

hellasol (approved):

PR #656, opened 2022-11-17
improve handling for filler words at sentence start

### LINKED ISSUE #653: remove comma when proposing to remove filler words at the start of the sentence
*Really, a* lot of man\-hours have gone into the development\.

<p></p>

we should replace "Really, a" with "A" \.\. but this raises the question if we then highlight only "Really" but replace more?

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/653

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2458,17 +2458,25 @@ def style_word_analysis_de(
             if not is_sub_category_enabled(version, config, subcategory):
                 continue
 
-            if not is_word_match(token, tokens, lang, word, word_types, None, False):
+            if not is_word_match(token, tokens, lang, word, word_types):
                 continue
 
             alternatives = alternatives_declension(token, lang, alternatives)
+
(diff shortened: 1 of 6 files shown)


app/main.py
```diff
@@ -2458,17 +2458,25 @@ def style_word_analysis_de(
             if not is_sub_category_enabled(version, config, subcategory):
                 continue
 
-            if not is_word_match(token, tokens, lang, word, word_types, None, False):
+            if not is_word_match(token, tokens, lang, word, word_types):
```
lsmith77, line 2461: we disabled lower casing here which I now removed. the test suite did not expose any issues that would come as a result

lsmith77:
https://github.com/Azure/docker-login/issues/46

app/main.py
lsmith77: @lnazarenko ^

lnazarenko (approved):

PR #660, opened 2022-11-18
slack has no alternative limit

lsmith77:


app/main.py
```diff
@@ -134,6 +134,7 @@ async def handle_command_witty(
             user_request_in, settings.slack_organization_id
         )
 
+    user_request_in.config.__setattr__("alternatives_max_count", None)
```
lsmith77, line 137: here we force the limit to be removed

app/main.py
```diff
@@ -764,6 +765,7 @@ def apply_rules(user_request_in: RequestIn, configs: dict, plan: str):
 
     user_request_in.config.__setattr__("disabled_categories", disabled_categories)
     user_request_in.config.__setattr__("plan", plan)
+    user_request_in.config.__setattr__("alternatives_max_count", 5)
```
lsmith77: for now we force the value of 5

hellasol (approved):

PR #664, opened 2022-11-23
made m/f rule more generic

lsmith77:


tests/test_general_cases/test_api_english_gender_acronyms/output.json
```diff
@@ -8,16 +8,13 @@
...
                 {
-                    "text": "*/w/m"
+                    "text": "*/f/m"
                 },
                 {
-                    "text": "w/d/m"
-                },
-                {
-                    "text": "w/*/m"
+                    "text": "EEO"
```
lsmith77: @tje-ww accepting this alternative would then result in `Developer m/f` to become `Developer EEO` which imho doesn't make much sense. what was your idea here in regards to `EEO` ?

app/rules.py
```diff
@@ -749,8 +710,7 @@
     "push the boundaries",
     "fossil promo",
     "good friends",
-    "good at"
-    "of a kind",
+    "good at" "of a kind",
```
lnazarenko: missing comma  between "good at”  and "of a kind"

tests/test_general_cases/test_api_german_gender_acronyms_with_parenthesis/output.json
```diff
@@ -8,20 +8,14 @@
                     "remove": true
                 },
                 {
-                    "text": "(d/w/m)"
+                    "text": "(*/M/W)"
                 },
```
lnazarenko, line 12: (*/M/W) should be W replaced with F

lnazarenko (approved):

lsmith77:
@tje-ww the text also doesn't work well if `w` or `f` is the first letter

image

PR #674, opened 2022-11-27
ensure flexion is correct in german_noun_analysis

lsmith77:


app/main.py
```diff
@@ -1857,37 +1858,48 @@ def german_noun_analysis(word, genus_only=False):
...
-        while i < len(word) - 4:
-            partial_word = word[i:].capitalize()
-            i += 1
 
-            result = german_nouns[partial_word]
+        # skip the last 2 letters
+        while i < len(word) - 2:
+            partial_word = word[i:]
+
+            # avoid cases like 'Ende' at the end of 'Arbeitgebende'
```
lsmith77: @tje-ww any advice here? without this, `Arbeitgebende` will be reduced to `Ende` which leads to determining the wrong genus. are there other such cases you are aware of?

app/main.py
```diff
@@ -1840,59 +1842,87 @@ def word_types_overlap(a_word_types, b_word_types):
...
+
+    if "genus" in result:
+        return result
+
+    if "genus 1" in result:
+        result["genus"] = result["genus 1"]
+
+        return result
+
+    if word[-5:].lower() == "leute":
```
lsmith77, line 1986: edge case fix

app/main.py
```diff
@@ -1840,59 +1842,87 @@ def word_types_overlap(a_word_types, b_word_types):
...
-            if result != None:
-                return result
+        result = german_noun_lookup(partial_word.capitalize())
+        if result == None:
+            i += 1
+            continue
 
-    if isinstance(result, list) and "genus 1" in result:
-        result["genus"] = result["genus 1"]
+        result["Lemma"] = word
```
lsmith77: ensure we don't return the data for the partial result

app/rules.py
```diff
@@ -424,6 +424,39 @@
     "aber",
 ]
 
+rules["de"]["gender_neutral_nouns"] = {
+    "Ierende": {
```
lsmith77, line 427: this is "hack" to handle gender-neutral words like "Studierende", "Rotierende", "Levitierende" etc.

app/rules.py
```diff
@@ -424,6 +424,39 @@
...
+            "dativ singular": "Ierender",
+            "dativ plural": "Ierenden",
+            "akkusativ singular": "Ierende",
+            "akkusativ plural": "Ierende",
+        },
+        "lemma": "Ierende",
+        "pos": ["Substantiv", "adjektivische Deklination"],
+        "genus": "f",
+    },
+    "Gebende": {
```
lsmith77, line 442: handle cases like "Arbeitgebende"

lnazarenko (approved):

PR #711, opened 2023-01-16
ensure grammatical correct alternatives with 'zu' in a verb

### LINKED ISSUE #710: "sicherzustellen" ie. zu
Wir haben noch viel sicherzustellen\.

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/710

image


app/main.py
```diff
@@ -2341,7 +2341,34 @@ def align_verb_form(lang, a_token, b_token):
 
         return b_text
     elif lang.lang == "de":
-        ending = a_token.text[len(a_token.lemma_) :]
+        # check if "zu" was stripped from the word in the lemma
+        if "zu" in a_token.text and "zu" not in a_token.lemma_:
+            # find verb at the end
+            if (
+                not b_token.text.startswith("be")
```
lsmith77: does it make sense to skip all words with `be` (`bewirken`), `er` (`ermöglichen`) and `ge` (`gewährleisten`) at the start? are there other relevant prefixes?

lsmith77:
tried various splitter libs but they don't seem to be able to split up words like `austauschen`:

* https://github.com/dtuggener/CharSplit
* https://github.com/TimKam/compound-word-splitter
* https://github.com/karlb/wikdict-compound
* https://github.com/mammothb/symspellpy

Didn't try out

* https://github.com/repodiac/german_compound_splitter (CC license and unclear data source license)
*

app/main.py
tje-ww: Prefixes of verbs that can split with “zu”
ab-
an-
auf-
aus-
bei-
ein-
mit-
nach-
weg-
zu-
her-
nach-


For these prefixes, it depends on the verb, whether you can split it with zu or not:
um-
über-
durch-
unter-
um-

For these cases, maybe a verb list would help? https://www.verblisten.de/listen/verben/anfangsbuchstabe/zusammengesetzt/vollstaendig.html

tje-ww: You don’t split verbs with the prefixes ge-, er-, ent-, emp-, be-, hinter-, miss-, ver- and zer-.

lsmith77: ok thank you .. but what about `sicherzustellen` .. is this not correct? 
https://de.wiktionary.org/wiki/sicherzustellen

Based on what you said above this should be `zu sicherstellen` which is incorrect.

So should `sicher` be added to the list of `it depends`?

what about `hochstellen` => `hochzustellen` 

I think there is still a gap in the above rules.

tje-ww: Good catch, that’s another type of compound verb entirely - adjective + verb.
You can split them with “zu.” The big BUT: when you use the compound verb figuratively you spell it as one word, when you use it in the literal sense, you write separate words. 
Example
sicher(zu)stellen, das jemand in der Praxis ist
Opas alte Kristallkelche sicher (zu) stellen, damit sie nicht umfallen
Not sure, if we can catch that (yet). 

That said, there will be different spelling conventions (alte deutsche Rechtschreibung versus neue deutsche Rechtschreibung; Österreichische und Schweizerische Rechtschreibung) I’ll look that up when I get home (approx. 4:30 pm).

lnazarenko (approved):

PR #717, opened 2023-01-17
Pluralize singular they

### LINKED ISSUE #713: Change the verb form when choosing "they"
When I choose the alternative "they" and he or she gets replaced, the verb should be adapted to the right form\.

[Bildschirmaufnahme 2022\-06\-29 um 15\.21\.58\.mov](https://wittyworks.slack.com/files/UBN609G3H/F03N11VHSJD/bildschirmaufnahme_2022-06-29_um_15.21.58.mov)

<p></p>

he is \-\> they are

he has \-\> they have

he fixes, builds and repairs \-\> they fix, build and repair

<p></p>

<p></p>

<p></p>

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/713

note we also have other cases where we suggest `they` (or pluralize), we can look into leveraging this functionality there as well eventually.

@hellasol FYI this will lead to potentially a lot of text being highlighted. that being said, we might eventually optionally allow defining a shorter highlight range (and/or a way to also display a shorter alternative .. ie. `he` highlight with `they` as the alternative but behind the scenes it is `he is` that will be replaced by `they are`.

image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1829,10 +1829,9 @@ async def english_rules(
     if lang.locale == "en-GB":
         words_data_en["od"] = rules["en-GB"]["open_disc_words_data"]
         words_data_en["ge"] = rules["en-GB"]["gender_words_data"]
-        words_data_en["ge-singular-they"] = (
-            rules["en-GB"]["gender_words_data"]
-            + rules["en-GB"]["bias_singular_they_alternatives"]
-        )
+        words_data_en["ge-singular-they"] = rules["en-GB"][
+            "bias_singular_they_alternatives"
+        ]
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Install dependencies
         run: |
           pipenv install --deploy --dev --verbose
-          pipenv run python -m spacy download en_core_web_sm
+          pipenv run python -m spacy download en_core_web_md
```
lsmith77: without this change a lot of verbs were not detected properly.

lnazarenko (approved):

PR #737, opened 2023-01-31
a fossil -> an old person

### LINKED ISSUE #733: "such a fossil" => "such an old person"
<p></p>

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/733

@lnazarenko a better strategy for the recipes adding plural to singular etc might be inside the API. that way we would also know if we are going from singular to plural in an alternative.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2018,7 +2018,6 @@ async def english_rules(
         )
 
     if is_sub_category_enabled(version, config, "inclusive"):
-
         if is_sub_category_enabled(version, config, "d_and_i"):
             subcategory = "d_and_i"
             category = categories[subcategory]["category"]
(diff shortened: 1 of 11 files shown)


app/main.py
```diff
@@ -2394,10 +2394,11 @@ def align_noun_form(lang, a_text, a_token, b_token):
             if key in b_word["flexion"]:
                 return b_word["flexion"][key]
 
-        return b_text
-
     if b_token.morph.get("Number") == ["Sing"]:
         return Noun(b_text).plural()
+    # do we want to include plural alternatives for singular?
```
lsmith77: do we want to include plural alternatives for singular? or should we stop adding plural alternatives?

tests/test_sing_or_plur/test_plur/output.json
```diff
@@ -5,16 +5,16 @@
...
-                    "text": "Engineering team"
+                    "text": "Engineering teams"
                 },
                 {
-                    "text": "Development team"
+                    "text": "Development teams"
                 },
                 {
-                    "text": "Engineering"
+                    "text": "Engineerings"
```
lsmith77, line 14: how do we handle going from plural to singular alternative?

lnazarenko (approved):

PR #738, opened 2023-01-31
massdriver

lsmith77:
https://md.witty.works/


massdriver-lang/schema-artifacts.json
```diff
@@ -0,0 +1,145 @@
+{
```
likelymichael: These schema-*.json files are for local development, and don't need to be checked into a repo. They can be added to your `.gitignore` file.

massdriver-lang/src/_params.auto.tfvars.json
```diff
@@ -0,0 +1,55 @@
+{
```
likelymichael: The _connections*, _params*, and _md* files can also be added to your `.gitignore` file.

likelymichael (commented):
Added some context to files that can be added to `.gitignore` file so they're not checked into repo.

Dockerfile
```diff
@@ -1,45 +1,31 @@
...
-# and the start of gunicorn as wittyuser
-COPY ops/entrypoint.sh /entrypoint.sh
-RUN chmod +x /entrypoint.sh
-# app services need the port 2222 exposed for ssh access
-EXPOSE 2222
-
-CMD [ "/entrypoint.sh" ]
+
+RUN mkdir /code
+WORKDIR /code
```
likelymichael:
```suggestion

WORKDIR /code
```
`WORKDIR` will create the directory for you if it doesn't exist, so the `RUN mkdir` layer is not needed

Dockerfile
```diff
@@ -1,45 +1,31 @@
...
+
+COPY ./app /code/app
+
+COPY ./locales /code/locales
+
+COPY ./training_data /code/training_data
+
+# Creates a non-root user and adds permission to access the /code folder
+RUN adduser -u 5678 --disabled-password --gecos "" appuser && chown -R appuser /code
+USER appuser
```
likelymichael: If these are moved up to L20 above `WORKDIR`, then you wouldn't need to set folder permissions w/ `chown`

massdriver/src/main.tf
```diff
@@ -0,0 +1,8 @@
+module "helm" {
+  source             = "github.com/massdriver-cloud/terraform-modules//massdriver-application-helm?ref=b4401ac"
+  name               = var.md_metadata.name_prefix
+  namespace          = var.namespace
+  chart              = "${path.module}/chart"
+  kubernetes_cluster = var.kubernetes_cluster
+  additional_envs    = []
```
likelymichael:
```suggestion
```
We can remove this array, we aren't using this anymore (has been updated in our app template).

massdriver-lang/src/main.tf
```diff
@@ -0,0 +1,8 @@
+module "helm" {
+  source             = "github.com/massdriver-cloud/terraform-modules//massdriver-application-helm?ref=b4401ac"
+  name               = var.md_metadata.name_prefix
+  namespace          = var.namespace
+  chart              = "${path.module}/chart"
+  kubernetes_cluster = var.kubernetes_cluster
+  additional_envs    = []
```
likelymichael:
```suggestion
```
We can remove this array, we aren't supporting it anymore and it's been updated in our app templates.

likelymichael (changes requested):
Some formatting changes recommended for Dockerfile and helm chart

Dockerfile
```diff
@@ -1,45 +1,31 @@
...
+COPY requirements.txt requirements.txt
+RUN pip install --user -r requirements.txt
+
+COPY ./app /code/app
+
+COPY ./locales /code/locales
+
+COPY ./training_data /code/training_data
+
+# TODO compute WORKER = (CPU * 2 * MULTIPLIER) + 1
```
lsmith77: @chrisghill ideally we compute the worker count here based on the formula above. I guess this should happen inside the `massdriver.yaml` ?

chrisghill: Hey @lsmith77 , sorry I didn't see this until now.

If the worker count can be set via an environment variable, then I would recommend doing it in `massdriver.yaml`.  It's up to you whether you want to specify a safe default in the Dockerfile (`ENV WORKERS 4`). The only benefit here would be if you wanted your image to be able to be run without *requiring* setting the `WORKERS` env.

Environment variables set at runtime will override values set in a Dockerfile., so in `massdriver.yaml` you could have

```
app:
  envs:
    WORKERS: (.params.resourceRequests.cpu * 2 * .params.configuration.multiplier) | ceil | tostring
```

This obviously requires you have a field named `multiplier` in your `params.configuration` block.  The `ceil` and `tostring` are important to make sure you get an integer, and convert it into a string since all environment variables need to be strings.

lsmith77: @chrisghill ok great thx. is there some way to view the computed final environment variable values?

lsmith77: here is my attempt at implementing your suggestion https://github.com/witty-works/nlp_api/pull/738/commits/8e4c4b15d343fdc0492710ecde624939b8c111a0

chrisghill: You can use `kubectl` to `describe` either a pod or the deployment. This should give you a list of the environment variables.  For example:

```
$ kubectl describe pod nlpapi-main-nlpapi-5t4h-6f9659f9d8-fv544
Name:         nlpapi-main-nlpapi-5t4h-6f9659f9d8-fv544
Namespace:    default
...
    Environment:
      LANGS:                      ["en","de"]
      LOGGING_CONFIG_FILENAME:    stdout
      LOGGING_ENABLED:            true
      REDIS_PORT:                 6380
...
```

I pruned a bulk of the values (particularly sensitive values), but that gives you an idea.

.github/workflows/deployAzure.yaml
```diff
@@ -1,25 +1,30 @@
-on: [push]
-name: Build a Docker image and Push it to ACR
+name: Deploy to Massdriver
+on:
+  push:
+    branches: [feature/massdriver]
```
likelymichael: When this gets merged to `main`, this trigger should be updated to where you want it to be. Do you have a `dev` branch as well? If so, creating 2 GitHub actions (1 for prod, 1 for staging) is a great way to test changes in staging before merging to prod.

.github/workflows/deployAzure.yaml
```diff
@@ -29,13 +34,35 @@ jobs:
...
+
+      - name: Publish Bundle
+        uses: massdriver-cloud/actions/bundle_publish@v3
+        with:
+          build-directory: ./massdriver
+
+      - name: Set Image Version
+        uses: massdriver-cloud/actions/app_patch@v3
+        with:
+          project: nlpapi
```
likelymichael: If you do set up 2 GitHub actions, the `Set Image Version` and `Deploy App` steps will need to be updated to match the values for your other environment.

Dockerfile
```diff
@@ -1,43 +1,31 @@
...
+USER appuser
+
+ENV PATH "$PATH:/home/appuser/.local/bin"
+
+COPY requirements.txt requirements.txt
+RUN pip install --user -r requirements.txt
+
+COPY ./app /code/app
+
+COPY ./locales /code/locales
```
likelymichael: No `locales` directory in repo.

likelymichael (approved):
Other than the `locales` directory needing to be removed from the `Dockerfile`, the rest looks great! Left some informational comments about GitHub actions and how they can be utilized for different branches/environments.

.github/workflows/deployAzure.yaml
lsmith77: yeah .. we do have `dev`

isn't it possible to have a single GA to prevent differences between the two to cause problems.

shouldn't changing `target: main` to `target: github.ref_name` do the trick? that being said .. for prod releases I would technically prefer to use a git tag and not the `main` branch

@chrisghill do you have some advice here?

Dockerfile
```diff
@@ -1,43 +1,29 @@
...
+
+# change user https://stackoverflow.com/questions/73174175/gunicorn-command-not-found
+USER appuser
+
+COPY ./app /code/app
+
+COPY ./training_data /code/training_data
+
+ARG WORKERS=1
+CMD gunicorn app.main:app --preload -b 0.0.0.0:8081 -w 1 -k uvicorn.workers.UvicornWorker --forwarded-allow-ips "*"
```
lsmith77: @chrisghill if I use `-w $WORKERS` here I end up getting

```
usage: gunicorn [OPTIONS] [APP_MODULE]
gunicorn: error: argument -w/--workers: expected one argument
```

.github/workflows/deployAzure.yaml
likelymichael: > shouldn't changing target: main to target: github.ref_name do the trick? that being said .. for prod releases I would technically prefer to use a git tag and not the main branch

the `target:main` on lines 58 and 67 reference your application on massdriver. the `project`, `target`, and `manifest` fields merge together to create the slug used to designate which application to publish/deploy.

> isn't it possible to have a single GA to prevent differences between the two to cause problems.

you could set conditionals within a single github action to run `patch`/`publish`/`deploy` on all of your apps, but it could become daunting to manage as you scale

chrisghill: GHA does support using conditionals (https://docs.github.com/en/actions/learn-github-actions/expressions), however I find that writing, reading and managing a single workflow file with multiple conditionals for different environments is much more overhead than simply maintaining 2 separate workflow files. I've always maintained a separate file for each environment.

As Michael mentioned, the `target` field is referencing the Massdriver "environment" (previously called "target"). 
image

Dockerfile
chrisghill: So in a Dockerfile, [an `ARG` is an "internal" variable used for image building purposes](https://docs.docker.com/engine/reference/builder/#arg), it doesn't persist beyond the building process.  For example, we do this in one of our images:

```
ARG BASE_IMG=golang:1.18

FROM ${BASE_IMG} as base
...
```

This basically makes the image a "build-time variable" that can easily be changed at the top of the file.  You'll noticed it's referenced with `${...}`.  So if you did:

```
ARG WORKERS=1
CMD gunicorn app.main:app --preload -b 0.0.0.0:8081 -w ${WORKERS} -k uvicorn.workers.UvicornWorker --forwarded-allow-ips "*"
```

It should work, but this still isn't what you want. You're effectively "hardcoding" the number of workers in the image. You want to variable-ize it via an environment variable.

```
ENV WORKERS=1
CMD gunicorn app.main:app --preload -b 0.0.0.0:8081 -w $WORKERS -k uvicorn.workers.UvicornWorker --forwarded-allow-ips "*"
```

This will make it so the number of workers is determined by the `WORKERS` environment variable, which is defaulted to `1`, but can be overridden at image run time by setting that environment variable to something else.

PR #745, opened 2023-02-18
support postfix word matches

### LINKED ISSUE #743: prefix/regular expression rules
Is there a way for us to detect no\-go words in German compounds? I'm not looking to provide alternatives for these but rec' deleting them\. Examples:

Juden\-  
Neger\-

\-deutscher

<p></p>

<https://wittyworks.slack.com/archives/C02T0CSRZNV/p1676207625414509?thread_ts=1676205151.624769&channel=C02T0CSRZNV&message_ts=1676207625.414509>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/743

Remaining issues:

1. should postfix checking be a rule level flag (f.e. `Jude` should probably not be a rule itself - then again `Du Jude` is likely also a derogatory - but if used as a postfix - in front of a derogatory - it should be highlighted) or where the alternatives would be different if its a postfix (f.e. with the N-word where we also offer `Schwarz~e~/~Schwarzer ..`), then again for `postfix` on `openly_discriminating` we could also default the alternatives to `[-]` as currently implemented if the word is contained inside the token (with special handling for `Ische` to avoid matching `labyrinthische`)
2. how to deal with words like `Facharzt`, where the female version doesn't contain the male version ie. `a` vs. `ä` .. this requires a smarter algorithm for the alternatives
3. how to deal with words where the plural doesn't start with the male version and where spacy isn't able to determine the lemma like `Fachärzte` and even then we would need the plural version?

* for 1. it would just mean adding a column in the google sheet and adjusting the csv files accordingly
* for 2. maybe the list of problematic words isn't too long, so we could potentially go with the hacky solution currently implemented
* for 3. we might just need the plural version of each word (ie. `Arzt` and `Ärzte`, `Jude` and `Juden`) or could we do this somehow on the lemmatizer? or we look up the plural as currently implemented?

todo items:
* [ ] remove all the superflous titles and openly discriminating words (ie. we don't need `Betriebsarzt` or `Judenei` anymore
* [ ] add generic professions (like `Arzt`, `Konditor`) under `base_titles`


app/main.py
```diff
@@ -3159,59 +3173,73 @@ def gendered_denom_analysis_de(
...
-                    continue
+            match = is_word_match(
+                lang.lang,
+                tokens[i],
+                tokens,
+                word,
+                word_types,
+                None,
+                True,
+                subcategory == "titles",
```
lsmith77: this should be a rule level flag

app/main.py
```diff
@@ -3445,9 +3473,18 @@ def rules_based_words_phrase_matcher(
...
-            ):
+            match = is_word_match(
+                lang.lang,
+                token,
+                tokens,
+                word,
+                word_types,
+                matches_false,
+                True,
+                category == "openly_discriminating",
```
lsmith77: this should be a rule level flag

app/main.py
```diff
@@ -3159,59 +3171,87 @@ def gendered_denom_analysis_de(
...
+                category,
+                subcategory,
+            )
+
+            if match == "postfix":
+                alternatives = alternatives.copy()
+                for k, alternative in enumerate(alternatives):
+                    replacement = tokens[i].lemma_
+                    alternative = alternative.replace(word, replacement)
+                    if word[0] == "A":
```
lsmith77, line 3380: handle `Arzt` vs. `Ärztin` and `ärztliche`

training_data/de/gendered_noun_words.csv
```diff
@@ -1,4 +1,5 @@
 Lemma,Word_Type,Category,Primary_subcategory,Sg_all_split,Pl_all_split,Alt_split
+Arzt,s,gendered,titles,"['ärztliche Untersuchung', 'Ärztin~Arzt']","['Ärzt~innen~ und ~Ärzte', 'ärztliche Untersuchung']","['ärztliche Untersuchung', 'Ärztin~Arzt', 'Ärzt~innen~ und ~Ärzte']"
```
lsmith77: rule missing in google sheet

app/main.py
```diff
@@ -3445,10 +3509,28 @@ def rules_based_words_phrase_matcher(
...
+                tokens,
+                word,
+                word_types,
+                matches_false,
+            )
+
+            if not match:
+                if (
+                    lang.lang == "en"
+                    or word == "Ische"
```
lsmith77: avoid match on `labyrintische`

training_data/de/gendered_noun_words.csv
```diff
@@ -299,3 +297,5 @@ Zuhörer,s,gendered,function,"['Zuhörer~in~/~Zuhörer', 'Publikum', 'Publikumss
 Zulieferer,s,gendered,function,"['Zulieferung', 'Zulieferfirma', 'zulieferndes Unternehmen', 'Zulieferunternehmen', 'Zuliefer~in~/~Zulieferer', 'Zuliefernde']",['Zuliefernde'],"['Zulieferung', 'Zulieferfirma', 'zulieferndes Unternehmen', 'Zulieferunternehmen', 'Zuliefer~in~/~Zulieferer', 'Zuliefernde']"
 anwenderbezogen,s,gendered,hidden_image,['praxisbezogen'],[''],['praxisbezogen']
 jedermann,s,gendered,hidden_image,"['Alle', 'alle miteinander', 'allerseits', 'Durchschnittsmensch', 'Jed~e~/~Jeder', 'alle', 'Durchschnittsmenschen', 'alle miteinander', 'allerseits']","['alle', 'Durchschnittsmenschen', 'alle miteinander', 'allerseits']","['Alle', 'alle miteinander', 'allerseits', 'Durchschnittsmensch', 'Jed~e~/~Jeder', 'alle', 'Durchschnittsmenschen', 'alle miteinander', 'allerseits']"
+Ingenieur,s,gendered,base_titles,"['===technik', 'Ingenieur~in~/~Ingenieur']","['Ingenieur~innen~ und ~Ingenieur']","['Technik', 'Ingenieur~in~/~Ingenieur', 'Ingenieur~innen~ und ~Ingenieur']"
```
lsmith77: `===` means put the prefix here .. ie. `Elektrotechnik` for `Elektroingenieur`

lsmith77:
merged into https://github.com/witty-works/nlp_api/pull/797 .. disabled the openly_discriminating logic for now

PR #781, opened 2023-03-28
integrate setfit context aware checker

### LINKED ISSUE #770: move azureml based API to the NLP API repo
goals

- improving performance \(no HTTP requests\)

- scalability

- reducing complexity \(only one system to monitor\)

- move the code to GIT

- reduce costs, especially since right now most requests do not need the context aware check but we need to run enough instances to be able to handle multiple calls in short succession

steps

- determine where to train models \(keep it on azure or move it else where?\)

- move the training model into the NLP API``

```

import os
import logging
import json
import transformers
from setfit import SetFitModel



def init():
    """
    This function is called when the container is initialized/started, typically after create/update of the deployment.
    You can write the logic here to perform init operations like caching the model in memory
    """
    global model
    # AZUREML_MODEL_DIR is an environment variable created during deployment.
    # It is the path to the model folder (./azureml-models/$MODEL_NAME/$VERSION)
    print(os.getenv("AZUREML_MODEL_DIR"))
    model = SetFitModel.from_pretrained(os.path.join(os.getenv("AZUREML_MODEL_DIR"), "outputs"))
    logging.info("Init complete")


def run(raw_data):
    """
    This function is called for every invocation of the endpoint to perform the actual scoring/prediction.
    In the example we extract the data from the json input and call the scikit-learn model's predict()
    method and return the result back
    """
    logging.info("Request received")
    data = json.loads(raw_data)["data"]

    result = model([data])
    logging.info("Request processed")
    return str(result[0])
```

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/770

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -8,12 +8,15 @@ import uvicorn
 import json
 import secrets
 import aiohttp
+import requests
 from typing import Optional, Union, List
 from collections import defaultdict
 from pydantic import parse_obj_as
 
 import os
 import fasttext
+
+from setfit import SetFitModel
 
 from spacy.matcher import PhraseMatcher, Matcher
 import pandas as pd
@@ -79,7 +82,7 @@ from app.models import (
     ErrorMessage,
     PrettyJSONResponse,
 )
-from app.lang_detection import get_lang_detection
+from app.lang_detection import LangDetection
 from app.categories import categories
 from app.settings import get_settings
 from app.logger import set_up_logger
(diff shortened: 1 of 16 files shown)


app/main.py
```diff
@@ -1708,19 +1687,26 @@ def is_sub_category_enabled(version: float, config: Config, subcategory: str):
...
+    for sentence in tokens.sents:
+        for i in range(start, sentence.end_char):
+            sentences[i] = sentence
+
+        start = sentence.end_char
+
+    for result in list_results:
+        words = result.text.lower().split()
+        if words[-1] in rules[lang]["context_check"] and result.end in sentences:
+            context_valid = setfit_model(sentence.text)
```
lsmith77: @fgbelidji

lsmith77:
`To ensure your model is fully loaded you may want to load then inside the "on_starting" hook and wait inside this hook until the models are loaded. When used with preload this should be enough. `

lsmith77:
* https://github.com/benoitc/gunicorn/issues/2478
* https://github.com/benoitc/gunicorn/issues/2608

lsmith77:
https://wittyworks.slack.com/archives/C02BVPTTQKE/p1681820153592169

PR #819, opened 2023-06-02
batch context aware checker calls

### LINKED ISSUE #807: batch context aware false positive checks

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/807

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1829,32 +1829,6 @@ def apply_false_positives(
     return list_results
 
 
-async def call_context_checker(sentences, result: ResultOut):
-    headers = {
-        "Content-Type": "application/json",
-        "Authorization": ("Bearer " + settings.context_checker_api_key),
-    }
-
-    sentence = None
-    for end_char in sentences:
-        if result.end <= end_char:
-            sentence = sentences[end_char]
-            break
-
-    if sentence == None:
-        return True
-
-    payload = {
-        "data": sentence.text,
-    }
-
-    context_valid = await fetch_json_post(
-        settings.context_checker_url, json.dumps(payload), headers, "context checker"
-    )
-
-    return context_valid == "1"
-
-
 def is_sub_category_enabled(config: Config, subcategory: str):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1874,23 +1848,68 @@ def is_sub_category_enabled(config: Config, subcategory: str):
...
+    )
+
+    keys_to_remove = []
+    for i in range(len(sentences)):
+        sentence = sentences[i]
+        if context_results[i] != "1":
+            for result_key in sentences_to_check[sentence]:
+                keys_to_remove.append(result_key)
+
+    keys_to_remove.sort(reverse=True)
```
lsmith77, line 1916: we need to remove from the end so that the list indexes so not change

lsmith77:
@lnazarenko FYI I already switched production to the new score.py file

lsmith77:
```
import os
import logging
import json
import transformers
from setfit import SetFitModel


def init():
    """
    This function is called when the container is initialized/started, typically after create/update of the deployment.
    You can write the logic here to perform init operations like caching the model in memory
    """
    global model
    # AZUREML_MODEL_DIR is an environment variable created during deployment.
    # It is the path to the model folder (./azureml-models/$MODEL_NAME/$VERSION)
    print(os.getenv("AZUREML_MODEL_DIR"))
    model = SetFitModel.from_pretrained(
        os.path.join(os.getenv("AZUREML_MODEL_DIR"), "outputs")
    )
    logging.info("Init complete")


def run(raw_data):
    """
    This function is called for every invocation of the endpoint to perform the actual scoring/prediction.
    In the example we extract the data from the json input and call the scikit-learn model's predict()
    method and return the result back
    """
    logging.info("Request received")
    data = json.loads(raw_data)["data"]

    if isinstance(data, list):
        result = model(data)
        result = list(map(str, result))
    else:
        result = model([data])
        result = str(result[0])

    logging.info("Request processed")

    return result
```

lnazarenko (approved):

PR #824, opened 2023-06-05
use lemma_lookup data to determine incorrect plural detection cases

lsmith77:


app/model.py
```diff
@@ -87,13 +77,133 @@ def custom_tokenizer(lang, nlp):
     )
 
 
-@German.factory("custom_lemmatizer_de")
-def custom_lemmatizer_de(nlp, name):
-    lemma_lookup = {
+lemma_plural_lookup = {
```
lsmith77, line 80: these are all the cases where the plural is clearly identifiable

app/model.py
```diff
@@ -87,13 +77,133 @@ def custom_tokenizer(lang, nlp):
...
+        "Learnings": "Learnings",
+        "Dreads": "Dreads",
+        "Charts": "Charts",
+    },
+    "en": {},
+}
+
+lemma_lookup = {
+    "de": {
+        "Behinderte": "Behinderte",
```
lsmith77, line 98: here the plural cannot be clearly identified due to overlap with the female singular form

lnazarenko (approved):

PR #836, opened 2023-06-29
revision of German lookup table

### LINKED ISSUE #825: re-generate lemma lookup
<https://github.com/witty-works/nlp_api/blob/dev/app/model.py#L91>

document how to do it
make sure we map to an a proper lemma since currently we sometimes map to a plural

lnazarenko:
- remove duplicates, 
- update german lookup table

fixes https://github.com/witty-works/nlp_api/issues/825


lsmith77:
@lnazarenko did you also document how you generated the lookup table? also do we need to update it as well for English?

lnazarenko:
I am writing the description how to update lookup table now.  It seems the English lookup table should be also updated because we currently use bigger model, please create a ticket for it.

PR #843, opened 2023-07-10
add german context checking model

### LINKED ISSUE #762: context checker for german

lsmith77:
fixes #762


app/main.py
```diff
@@ -1803,23 +1804,14 @@ def is_sub_category_enabled(config: Config, subcategory: str):
...
     sentences_to_check = defaultdict(list)
     for i in range(len(list_results)):
         result = list_results[i]
-        words = result.text.lower().split()
-        if not len(words):
-            continue
-
-        # a fossil => fossil
-        if words[-1] in rules[lang]["context_check"]:
+        if result.lemma in rules[lang]["context_check"]:
```
lsmith77, line 1814: we need to examine the lemma since otherwise the lookup fails if we have `unabhängige`

lsmith77: this also means we no longer need the hack for `a fossil` since the lemma would then still be `fossil`

PR #879, opened 2023-09-12
add ability to authenticate against Azure AD B2C and Office SSO + verify signature

### LINKED ISSUE #877: MS SSO
Word Plugin:

- Add fetching the Office Access Token

- Ensure MS SSO users is setup in the NLP API \(via the Dashboard\)

<p></p>

NLP API

- Ability to handle MS SSO access tokens

<p></p>

Dashboard

- Handle user creation for users that do not exist yet

<p></p>

Big question:

How can we reliably ensure mapping users from MS SSO to the Dashboard

<p></p>

<https://learn.microsoft.com/en-us/office/dev/add-ins/develop/sso-in-office-add-ins>

<p></p>

lsmith77:
fixes https://github.com/witty-works/nlp_api/issues/877


lsmith77:
https://medium.com/medialesson/verify-signature-of-aad-b2c-jwt-tokens-in-jwt-io-63222a69ece6

.env.example
```diff
@@ -0,0 +1,61 @@
+OVERWRITE_ENABLED_USER_CATEGORIES=False
```
hellasol, line 1: is none of this information sensitive?

app/auth_service.py
```diff
@@ -0,0 +1,169 @@
...
+    b2c_domain_name,
+):
+    _issuer = f"https://{b2c_domain_name}.b2clogin.com/{tenant_id}/v2.0/".lower()
+
+    try:
+        jwt.decode(
+            token, rsa_key, algorithms=["RS256"], audience=client_id, issuer=_issuer
+        )
+    except jwt.ExpiredSignatureError:
+        raise AuthError("Token error: The token has expired", 401)
```
hellasol, line 85: On the browser extension we do the refresh token call when a 403 is returned. So this will have to be updated

hellasol (approved):

.env.example
lsmith77: good question!

I doubled checked again and all of this is public information.

app/auth_service.py
lsmith77: yes we wrap the call to this function into our own try/except that converts it to a 403

PR #893, opened 2023-10-23
integrate languagetool

lsmith77:
The plan is to drop the languagetool platform.sh project to save money. As a result we will also archive the languagetool repository.

This also means that after merging this PR, we will need to update https://github.com/witty-works/nlp_api/blob/feature/integrate-languagetool/.github/workflows/main.yaml#L64 to point to the NLP API dev URL with a `lt.` prefix.

Once this is all merged to production, we will also need to make sure that https://lt.default.api.witty.works is still operational. The DNS should be cleaned up but in principle the DNS today already points to the correct region, so in theory it could just work once the current languagetool project is shut down.


lsmith77:
https://console.platform.sh/witty-works/2ycn7tlsdvd2o/feature%2Fintegrate-languagetool

https://feature-integrate-sdx5lua-2ycn7tlsdvd2o.fr-4.platformsh.site/docs#/default/post_debug_check_debug_check_post

hellasol (approved):

PR #953, opened 2023-12-21
Release/2.0.0

lsmith77:


hellasol (approved):
its a lot of changes.. maybe we could look at it together?

lsmith77:
yeah but we have reviewed most of this already multiple times. this is basically the big refactoring to use SQLite instead of CSV's

PR #962, opened 2024-01-03
update sqlite dump, check source declension only once

lsmith77:
also improve logging level handling, use models for lang/locale


tests/test_grammatically_correct_alternatives/test_german_en/output.json
```diff
@@ -51,7 +51,7 @@
                     "text": "wirkungsvollsten"
                 },
                 {
-                    "text": "als positiv empfunden"
```
lsmith77, line 54: this is now marked as an inspiration and is therefore not shown

hellasol (approved):

PR #995, opened 2024-02-05
as python-json is unmaintained, switch to pyjwt

lsmith77:
https://github.com/mpdavis/python-jose/issues/341


hellasol (commented):
look good to me, but cant approve as it is already merged

PR #1022, opened 2024-03-12
only check for highlights if the user is on a plan

lsmith77:


hellasol (approved):
@lsmith77 does this mean we wont show grey highlights anymore? If yes, can you open a tricket to remove this from the browser extension?

lsmith77:
yes .. will do

PR #1188, opened 2024-11-20
expand explanation to include long_text, video/image url

lsmith77:
fixes https://github.com/witty-works/browser-extension/issues/1357


app/models.py
```diff
@@ -853,7 +863,11 @@ def factory(
             if url is not None and len(url) == 0:
                 url = None
 
-            if url is not None and client.name == "web-ext":
+            if (
+                url is not None
+                and client.name == "web-ext"
+                and client.version < VersionString("1.40.0")
```
lsmith77, line 875: make sure this version is correct before releasing

PR #1193, opened 2024-11-27
move managing word type specific lemmatizations to the rule editor

lsmith77:


app/rules.py
```diff
@@ -562,11 +561,6 @@ def fetch_static_rules(langs: list[str]):
             static_rules[LangType.FR]["inclusive_articles"]
         )
 
-        static_rules[LangType.FR]["lemma_by_word_type"][WordType.ADJECTIVE] = {
```
lsmith77, line 565: need to be added in the rule editor

